PR #362, opened 2020-10-03
Add a contract for governance proposals

abarmat:
This PR introduce a contract to govern proposals passed by authorized proposers and approved by a multisig.

The proposals are IPFS hash links that can contain metadata in arbitratry format with the full description of the proposal.

### DIFF
--- a/cli/commands/deploy.ts
+++ b/cli/commands/deploy.ts
@@ -1,18 +1,63 @@
 import yargs, { Argv } from 'yargs'
 
-import { deployContract } from '../network'
+import {
+  getContractAt,
+  deployContract,
+  deployContractWithProxy,
+  deployContractAndSave,
+  deployContractWithProxyAndSave,
+} from '../network'
 import { loadEnv, CLIArgs, CLIEnvironment } from '../env'
 import { logger } from '../logging'
 
 export const deploy = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<void> => {
   const contractName = cliArgs.contract
   const initArgs = cliArgs.init
+  const deployType = cliArgs.type
+  const buildAcceptProxyTx = cliArgs.buildTx
 
-  logger.log(`Deploying contract ${contractName}...`)
+  // Get the GraphProxyAdmin to own the GraphProxy for this contract
+  const proxyAdminEntry = cli.addressBook.getEntry('GraphProxyAdmin')
+  if (!proxyAdminEntry) {
+    throw new Error('GraphProxyAdmin not detected in the config, must be deployed first!')
+  }
+  const proxyAdmin = getContractAt('GraphProxyAdmin', proxyAdminEntry.address)
 
   // Deploy contract
   const contractArgs = initArgs ? initArgs.split(',') : []
-  await deployContract(contractName, contractArgs, cli.wallet)
+  switch (deployType) {
+    case 'deploy':
+      logger.log(`Deploying contract ${contractName}...`)
+      await deployContract(contractName, contractArgs, cli.wallet)
+      break
+    case 'deploy-save':
+      logger.log(`Deploying contract ${contractName} and saving to address book...`)
+      await deployContractAndSave(contractName, contractArgs, cli.wallet, cli.addressBook)
+      break
+    case 'deploy-proxy':
+      logger.log(`Deploying contract ${contractName} with proxy ...`)
+      await deployContractWithProxy(
+        proxyAdmin,
+        contractName,
+        contractArgs,
+        cli.wallet,
+        true,
+        buildAcceptProxyTx,
+      )
+      break
+    case 'deploy-proxy-save':
+      logger.log(`Deploying contract ${contractName} with proxy and saving to address book...`)
+      await deployContractWithProxyAndSave(
+        contractName,
+        contractArgs,
+        cli.wallet,
+        cli.addressBook,
+        buildAcceptProxyTx,
+      )
+      break
+    default:
+      logger.error('Please provide the correct option for deploy type')
+  }
 }
 
 export const deployCommand = {
@@ -32,6 +77,17 @@ export const deployCommand = {
         type: 'string',
         requiresArg: true,
         demandOption: true,
+      })
+      .option('t', {
+        alias: 'type',
+        description: 'Choose deploy, deploy-save, deploy-proxy, deploy-proxy-save',
+        type: 'string',
+        requiresArg: true,
+        demandOption: true,
+      })
+      .option('b', {
+        alias: 'build-tx',
+        description: 'Build the acceptProxy tx and print it. Then use tx data with a multisig',
       })
   },
   handler: async (argv: CLIArgs): Promise<void> => {
--- a/cli/commands/deploy.ts
+++ b/cli/commands/deploy.ts
@@ -3,8 +3,8 @@ import yargs, { Argv } from 'yargs'
 import {
   getContractAt,
   deployContract,
-  deployContractWithProxy,
   deployContractAndSave,
+  deployContractWithProxy,
   deployContractWithProxyAndSave,
 } from '../network'
 import { loadEnv, CLIArgs, CLIEnvironment } from '../env'
@@ -27,7 +27,7 @@ export const deploy = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<voi
       logger.log(`Deploying contract ${contractName} and saving to address book...`)
       await deployContractAndSave(contractName, contractArgs, cli.wallet, cli.addressBook)
       break
-    case 'deploy-proxy':
+    case 'deploy-with-proxy':
       // Get the GraphProxyAdmin to own the GraphProxy for this contract
       const proxyAdminEntry = cli.addressBook.getEntry('GraphProxyAdmin')
       if (!proxyAdminEntry) {
@@ -41,11 +41,10 @@ export const deploy = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<voi
         contractName,
         contractArgs,
         cli.wallet,
-        true,
         buildAcceptProxyTx,
       )
       break
-    case 'deploy-proxy-save':
+    case 'deploy-with-proxy-save':
       logger.log(`Deploying contract ${contractName} with proxy and saving to address book...`)
       await deployContractWithProxyAndSave(
         contractName,
@@ -80,7 +79,7 @@ export const deployCommand = {
       })
       .option('t', {
         alias: 'type',
-        description: 'Choose deploy, deploy-save, deploy-proxy, deploy-proxy-save',
+        description: 'Choose deploy, deploy-save, deploy-with-proxy, deploy-with-proxy-save',
         type: 'string',
         requiresArg: true,
         demandOption: true,
@@ -88,6 +87,8 @@ export const deployCommand = {
       .option('b', {
         alias: 'build-tx',
         description: 'Build the acceptProxy tx and print it. Then use tx data with a multisig',
+        default: false,
+        type: 'boolean',
       })
   },
   handler: async (argv: CLIArgs): Promise<void> => {
--- a/cli/commands/deploy.ts
+++ b/cli/commands/deploy.ts
@@ -16,13 +16,6 @@ export const deploy = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<voi
   const deployType = cliArgs.type
   const buildAcceptProxyTx = cliArgs.buildTx
 
-  // Get the GraphProxyAdmin to own the GraphProxy for this contract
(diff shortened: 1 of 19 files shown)


abarmat:
We should update pragma to 0.7.3 like the rest of the contracts and review that it is compiling as the code is outdated.

davekaj (commented):
some comments we should address b4 merging

contracts/governance/GraphGovernance.sol
```diff
@@ -0,0 +1,84 @@
...
+        bytes32 _votesProof,
+        ProposalResolution _resolution
+    ) external override onlyGovernor {
+        require(_proposalId != 0x0, "!proposalId");
+        require(_votesProof != 0x0, "!votesProof");
+        require(_resolution != ProposalResolution.Null, "!resolved");
+        require(!isProposalCreated(_proposalId), "proposed");
+
+        proposals[_proposalId] = Proposal({ votesProof: _votesProof, resolution: _resolution });
+        emit ProposalCreated(msg.sender, _proposalId, _votesProof, _resolution);
```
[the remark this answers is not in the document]
abarmat: The community poll results should result in an IPFS hash that must be submitted in the votesProof field, while the `proposalId` is a bytes32 IPFS hash of the content of the proposal.

contracts/governance/GraphGovernance.sol
```diff
@@ -0,0 +1,84 @@
...
+    function isProposalCreated(bytes32 _proposalId) public view override returns (bool) {
+        return proposals[_proposalId].votesProof != 0;
+    }
+
+    /**
+     * @notice Submit a new proposal.
+     * @param _proposalId Proposal identifier. This is an IPFS hash to the content of the proposal
+     * @param _votesProof An IPFS hash of the collection of signatures for each vote
+     */
+    function createProposal(
```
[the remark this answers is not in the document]
abarmat: As it changed the focus to only creating a proposal after the fact (when it was already voted in Snapshot) with the purpose to inscribe, I think the multisig should create it so we don't allow a single member to inscribe proposals as valid.

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
davekaj: I was under the impression there is a vote where we tally the votes for the council, and a vote for the community. Both in snapshot. Maybe I am mistaken?

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
davekaj: I am okay with either. We can go with what you have done here, unless Brandon has a different opinion

davekaj:
Another question, where do we plan to emit documents such as the Arbitration Charter?

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
abarmat: Let's check that, I was thinking that the Council votes the same proposal with the rest of the community

abarmat:
> Another question, where do we plan to emit documents such as the Arbitration Charter?

It's a good question. We can publish in the same way as any other proposal using the IPFS hash of the document.

davekaj:
> > Another question, where do we plan to emit documents such as the Arbitration Charter?
> 
> It's a good question. We can publish in the same way as any other proposal using the IPFS hash of the document.

Lets discuss in our meeting today if we should fit it into the proposals functions, or create a unique function, such as `uploadDocument()` or something. I am making some open questions for the meeting today which we can use.

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
davekaj: yep we keep as is

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
davekaj: yes two different votes, resolved

davekaj:
@abarmat fix is ready for review. if good we can merge and test on testnet

abarmat (commented):
Let's remove `.only` from the Governance test.

I would add inside the "should create a proposal" test a validation to see if the proposal struct is properly stored, checking every field against what we send in createProposal.

test/governance/gov.test.ts
```diff
@@ -0,0 +1,131 @@
...
+const { ethers } = hre
+const { HashZero } = ethers.constants
+
+enum ProposalResolution {
+  Null,
+  Accepted,
+  Rejected,
+}
+
+describe.only('GraphGovernance', () => {
```
spalladino, line 19: We're using eslint [`no-only-tests`](https://www.npmjs.com/package/eslint-plugin-no-only-tests) to prevent against accidentally committing these - it happened a lot!

abarmat: Good advise :)

contracts/governance/GraphGovernance.sol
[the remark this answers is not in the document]
davekaj: To add more detail - yes there are two different votes, but we will only emit the council votes here. We will not add the community votes. The community vote should be obtainable by viewing the  proposal ID

contracts/governance/GraphGovernance.sol
```diff
@@ -0,0 +1,92 @@
...
+     * @return True if the proposal is already created
+     */
+    function isProposalCreated(bytes32 _proposalId) public view override returns (bool) {
+        return proposals[_proposalId].votes != 0;
+    }
+
+    /**
+     * @notice Submit a new proposal.
+     * @param _proposalId Proposal identifier. This is an IPFS hash to the content of the proposal
+     * @param _votes An IPFS hash of the collection of signatures for each vote
```
davekaj, line 52: a thought - brandon mentioned in our meeting yesterday that not all proposals will necessarily have a Snapshot vote (i.e. IPFS hash) associated with them. 

What I understand right now is that `createProposal` will always be linked to a  Graph Governance Proposal (GGP), and a GGP may or may not be related to a GIP, which may or may not be associated with a Snapshot IPFS hash

Ariel, wdyt? How do you understand it right now?

davekaj: also, a GGP can be created before or after a contract upgrade. in an emergency, the upgrade would happen before the GGP was sent out to the community

davekaj: Now i am thinking out loud, maybe this is how it works:
- a GGP is done on snapshot. This will always happen for an upgrade. Sometimes before, sometimes after. But it does not matter if it is before or after, as we are just storing governance proposals on chain.
- Therefore the proposal ID is the GGP ipfs hash
- Therefore the community votes, and the GIP ipfs hash, are not being uploaded here. If anything, the GGP will reference any necessary GIPs, and at GIP will mention any necessary snapshot votes 

WDYT ??

contracts/governance/GraphGovernanceStorage.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.7.3;
+
+import "./Governed.sol";
+import "./IGraphGovernance.sol";
+
+contract GraphGovernanceV1Storage is Governed {
```
davekaj, line 8: A design decision we are making is not including this Contract in the Controller. Which i think is the right choice.

But i would like to briefly discuss what we plan to do for upgrading the contract. This seems like the plan:
- It follows normal proxy pattern
- the owner is the graph council multisig
- we upgrade the code when needed.

The other option would be, starting a new contract from scratch, and stopping using the V1. I would prefer to use this naturally. BUT - because all of our other contracts use the proxy pattern. I am cool with keeping this as proxy. 

Sounds good? @abarmat  ?
Do you agree

contracts/governance/GraphGovernance.sol
spalladino: > a GGP is done on snapshot. This will always happen for an upgrade. Sometimes before, sometimes after. But it does not matter if it is before or after, as we are just storing governance proposals on chain.

Sorry for the noise, but in the mid-term this could happen in Defender, so the process of 1) setting up the tx, 2) collecting all council signatures, and 3) publishing it to IPFS, happens all within the same app.

contracts/governance/GraphGovernance.sol
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * @title Graph Governance Contract
+ * @notice Governance contract used to inscribe Graph Council and community votes.
+ */
+contract GraphGovernance is GraphGovernanceV1Storage, GraphUpgradeable, IGraphGovernance {
+    // -- Events --
+
+    event ProposalCreated(
+        address submitter,
```
davekaj, line 17: This will always be the governor - i.e. the graph council multisig. 

do we care to emit this every time?

contracts/governance/GraphGovernance.sol
davekaj: This seems ideal to me. For now we can get the council to work inside of both snapshot and defender. but in the future if defender could use a snapshot plugin that created proposals as well as voted on proposals, that would be great. it would make the council members tasks easier to handle

contracts/governance/GraphGovernance.sol
abarmat: Probably we can remove it, we can read who is the submitter by reading the contract governor

contracts/governance/GraphGovernanceStorage.sol
abarmat: The contract will be deployed like a pair (GraphProxy -> GraphGovernance) so we can change the implementation in the future. What we can do is to set the GraphProxy admin to our current GraphProxyAdmin contract so we can manage upgrades from the same place like the rest.

So the final setup would be:

**GraphProxyAdmin -> (owns) -> GraphProxy -> (uses) -> GraphGovernance**

contracts/governance/GraphGovernance.sol
davekaj: ill remove it

davekaj:
Hi Ariel, I believe we are good to go here. 

I tested on Rinkeby. If we are good we can probably deploy this to mainnet next week. wdyt?

abarmat:
@davekaj only thing missing is that we could set `proposalId` to indexed in the events

cli/commands/deploy.ts
```diff
@@ -1,18 +1,63 @@
...
 import { logger } from '../logging'
 
 export const deploy = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<void> => {
   const contractName = cliArgs.contract
   const initArgs = cliArgs.init
+  const deployType = cliArgs.type
+  const buildAcceptProxyTx = cliArgs.buildTx
 
-  logger.log(`Deploying contract ${contractName}...`)
+  // Get the GraphProxyAdmin to own the GraphProxy for this contract
```
abarmat, line 19: I think this logic only needs to happen if deployType is deploy-proxy

contracts/governance/GraphGovernance.sol
```diff
@@ -0,0 +1,110 @@
...
+        bytes32 votes,
+        bytes32 metadata,
+        ProposalResolution resolution
+    );
+
+    /**
+     * @notice Initialize this contract.
+     */
+    function initialize(address _governor) public onlyImpl {
+        require(_governor != address(0), "governor !0");
```
abarmat, line 39: What this the intended error message? or was it "!governor"

abarmat (commented):
@davekaj the integration tests are failing

contracts/governance/GraphGovernance.sol
davekaj: IIRC i was thinking `!governor` is more appropriate for when you are trying to access a function that only the governor can call. Maybe we can just name this error message to "Cannot set governor  = 0" or "governor != 0". 

Or , we can just go with `!governor` . What do you prefer?

cli/commands/deploy.ts
davekaj: yes i see what you are saying

contracts/governance/GraphGovernance.sol
davekaj: i went with `governor != 0`

abarmat: Makes sense

cli/network.ts
```diff
@@ -269,15 +288,17 @@ export const deployContractAndSave = async (
...
   })
+  logger.log('> Contract saved to address book')
 
   return deployResult.contract
 }
 
 export const deployContractWithProxyAndSave = async (
   name: string,
-  args: Array<{ name: string; value: string }>,
+  args: Array<any>,
```
abarmat, line 298: The issue with this being any is that if we pass an array with no parameter names then the entry will not be saved with the same format. Let me think about a refactor

davekaj (approved):
refactor looks good

PR #430, opened 2020-12-06
Cache address resolution from the Controller

abarmat:
This is saving between 5% to 15% gas depending on the transaction.

Related to: https://github.com/graphprotocol/contracts/issues/433

PR #434, opened 2020-12-09
Use a contract address cache for gas saving

abarmat:


abarmat:
Continued in https://github.com/graphprotocol/contracts/pull/430

PR #448, opened 2021-02-02
Merge in the v1.1.0 updates into master from testnet

davekaj:
We have a single package now, and thus, the newest version should always be on master going forwards


Jannis (approved):
LGTM

PR #450, opened 2021-02-15
Set indexer rewards destination

abarmat:
- Indexer can set a rewards destination for indexing rewards after closing an allocation and for rebate claims
- If rewards destination is set to zero, indexing rewards are restaked


contracts/staking/StakingStorage.sol
```diff
@@ -82,3 +82,7 @@ contract StakingV1Storage is Managed {
     // Allowed AssetHolders: assetHolder => is allowed
     mapping(address => bool) public assetHolders;
 }
+
+contract StakingV2Storage is StakingV1Storage {
```
pi0neerpat, line 86: Nice proxy skillz

contracts/staking/Staking.sol
```diff
@@ -1237,16 +1245,7 @@ contract Staking is StakingV1Storage, GraphUpgradeable, IStaking {
...
-            // Assign claimed tokens
-            if (restake) {
-                // Restake to place fees into the indexer stake
-                _stake(alloc.indexer, tokensToClaim);
-            } else {
-                // Transfer funds back to the indexer
-                require(graphToken.transfer(alloc.indexer, tokensToClaim), "!transfer");
-            }
-        }
+        _sendRewards(graphToken, tokensToClaim, alloc.indexer, restake);
```
pi0neerpat, line 1248: This means the indexer agent has to be updated too right? Since this is only for query fees, I think its ok to not have this feature in the agent for now

pi0neerpat (approved):
1. call `setRewardsDestination(address _destination)` from your indexer.
2. close allocation, which will call `_distributeRewards` and then `_sendRewards`

Clean and simple, I like it!

pi0neerpat:
Any tests planned to be added?

abarmat:
> Any tests planned to be added?

Yes, I'm pushing them in a minute

contracts/staking/Staking.sol
```diff
@@ -717,6 +717,14 @@ contract Staking is StakingV1Storage, GraphUpgradeable, IStaking {
         _withdraw(msg.sender);
     }
 
+    /**
+     * @dev Set the destination where to send rewards.
+     * @param _destination Rewards destination address. If set to zero, rewards will be restaked
+     */
+    function setRewardsDestination(address _destination) public override {
+        rewardsDestination[msg.sender] = _destination;
```
davekaj, line 729: Ariel slacked me saying he will add an event here, good idea

contracts/staking/StakingStorage.sol
```diff
@@ -82,3 +82,7 @@ contract StakingV1Storage is Managed {
...
+    mapping(address => address) public rewardsDestination;
```
davekaj, line 87: We could add a comment here to describe what `rewardsDestination` is

contracts/staking/Staking.sol
```diff
@@ -717,6 +717,14 @@ contract Staking is StakingV1Storage, GraphUpgradeable, IStaking {
         _withdraw(msg.sender);
     }
 
+    /**
+     * @dev Set the destination where to send rewards.
+     * @param _destination Rewards destination address. If set to zero, rewards will be restaked
```
davekaj, line 726: great idea that set to 0 = restaked. this way, when the upgrade happens, it will automatically be set so that everyone is re-staking. we won't run into any breaks

contracts/staking/Staking.sol
davekaj: Can we make it external? This is what the interface shows, seems we are no using it within the contract

contracts/staking/Staking.sol
```diff
@@ -1563,9 +1563,43 @@ contract Staking is StakingV1Storage, GraphUpgradeable, IStaking {
...
+
+        if (_restake) {
+            // Restake to place fees into the indexer stake
+            _stake(_beneficiary, _amount);
+        } else {
+            // Transfer funds to the beneficiary's designated rewards destination if set
+            address destination = rewardsDestination[_beneficiary];
+            require(
+                _graphToken.transfer(
+                    destination == address(0) ? _beneficiary : destination,
```
davekaj, line 1609: I believe this ternary is not needed. If we got this far, we already know that `rewardsDestination[_beneficiary] != 0`

this would work:
```
require(_graphToken.transfer(rewardsDestination[beneficiary]), _amount, "!transfer");
```
wdyt?

davekaj (changes requested):
In general looks good, looking forward to see the tests

contracts/staking/Staking.sol
davekaj: event added :)

test/rewards/rewards.test.ts
```diff
@@ -594,13 +596,71 @@ describe('Rewards', () => {
...
+        // NOTE: calculated manually on a spreadsheet
+        const expectedIndexerStake = beforeIndexer1Stake
+        const expectedTokenSupply = beforeTokenSupply.add(expectedIndexingRewards)
+        // Check stake should not have changed
+        expect(toRound(afterIndexer1Stake)).eq(toRound(expectedIndexerStake))
+        // Check indexing rewards are received by the rewards destination
+        expect(toRound(afterIndexer1Balance)).eq(
+          toRound(beforeIndexer1Balance.add(expectedIndexingRewards)),
+        )
+        // Check indexing rewards are sent from the staking contract
```
davekaj, line 662: is it more accurate to say `check indexing rewards were not sent to the staking contract` ? (basically they get minted and send directly to the indexer AFAICT)

test/rewards/rewards.test.ts
```diff
@@ -594,13 +596,71 @@ describe('Rewards', () => {
...
+        expect(toRound(event.amount)).eq(toRound(expectedIndexingRewards))
+
+        // After state
+        const afterTokenSupply = await grt.totalSupply()
+        const afterIndexer1Stake = await staking.getIndexerStakedTokens(indexer1.address)
+        const afterIndexer1Balance = await grt.balanceOf(indexer1.address)
+        const afterStakingBalance = await grt.balanceOf(staking.address)
+
+        // Check that rewards are put into indexer stake
+        // NOTE: calculated manually on a spreadsheet
```
davekaj, line 653: Note sure what this note really means

davekaj (commented):
Tests added look good!

contracts/staking/Staking.sol
abarmat: So to not change the current claim() interface, if you pass this method restake=true it will restake if restake=false it will send rewards out.

contracts/staking/Staking.sol
abarmat: I'll change it to be external. An event was added to this function too.

contracts/staking/StakingStorage.sol
abarmat: Good idea

contracts/staking/Staking.sol
abarmat: It depends, when the function is called from _claim() it might not be the case.

test/rewards/rewards.test.ts
abarmat: It is an outdated comment from the test I used as base for this one, I'll update it

test/rewards/rewards.test.ts
abarmat: Yes, what I tried to say is that rewards are out of the staking contract

abarmat:
I did some benchmark of gas usage for the solution that send rewards directly and the one that uses a rewards pool:

```
            |  Contract      ·  Method                 ·  Min        ·  Max        ·  Avg        ·  # calls    ·  usd (avg)   │
RewardPool: |  Staking       ·  closeAllocation        ·          -  ·          -  ·     256898  ·          1  ·           -  │
DirectTx:   |  Staking       ·  closeAllocation        ·          -  ·          -  ·     269114  ·          1  ·           -  │
```

Sending directly to the beneficiary on close is about 4% more expensive. So it might be a good idea to have a rewards pool where multiple rewards across many `closeAllocations()` and `claim()` are accumulated before calling `withdrawRewards()`, this will be only more efficient in the cases an indexer waits to call withdraw.

@davekaj wdyt?

contracts/staking/Staking.sol
davekaj: you are right ! there is nothing wrong with this code

contracts/staking/Staking.sol
davaymne: Will _destination be set up via Remix or some key in Agent config?

abarmat: `setRewardsDestination()` is a function that needs to be called by the indexer address (the one that staked), in the case of an indexer through a TokenLock it needs be called initially through Remix. The feature can be potentially integrated in the frontend too.

davekaj (approved):
changes look good to me

davekaj (approved):
Looks good to me. It would be nice to run some tests before actually merging to master.

which makes me think - what should our strategy be for merging to master?

My first thought is that we should only merge to master when the changes are actually deployed to mainnet. Meaning we could create a dev branch or something right now.

I am open to other suggestions!

pi0neerpat:
Just a suggestion, but you're welcome to copy some of what we do here https://github.com/superfluid-finance/protocol-monorepo

- Set up a canary branch ("main" or "dev")
- Set this as default branch when people land on the repo, or make any PRs
- Every time a PR is merged to master, the new contracts are deployed (upgraded) to testnet
- When things are upgraded/deployed to mainnet then we merge to "master" to reflect this as you mentioned @davekaj

PR #452, opened 2021-02-17
Accumulated subgraph rewards reset to zero on edge case

abarmat:
### Description
The RewardsManager contract main responsibility is to calculate rewards accrued for allocations based on the total token supply, tokens signalled on subgraphs, accumulated rewards on a subgraph and allocations. Under certain edge condition the calculation could fail and block closing an allocation that would collect rewards.

### Problem
There is a particular edge case that happens when there are running allocations on a subgraph deployment ID, the total amount of curation signal is removed back to zero and `closeAllocation()` is called by an indexer. If this happens, the indexer won't be able to close that allocation unless it is opting for no rewards by sending POI=0x0. The reason for this is that `getAccRewardsForSubgraph()` is reseting the accumulated rewards for a subgraph when the curation pool is empty. The consequence is that `subgraph.accRewardsForSubgraphSnapshot` can be larger than `subgraph.accRewardsForSubgraph` leading to a revert when calculating new accrued rewards.

### Solution
By removing the early check within `getAccRewardsForSubgraph()` that returns zero for the subgraph accumulated rewards  when the signal is zero we will always return the accumulated amount, as it should be an ever increasing number.

### Additional Changes
- Check that rewards > 0 before minting
- Linting differences
- Improve some comments

### DIFF
--- a/contracts/rewards/RewardsManager.sol
+++ b/contracts/rewards/RewardsManager.sol
@@ -187,9 +187,10 @@ contract RewardsManager is RewardsManagerV1Storage, GraphUpgradeable, IRewardsMa
         uint256 p = graphToken.totalSupply();
         uint256 a = p.mul(_pow(r, t, TOKEN_DECIMALS)).div(TOKEN_DECIMALS);
 
-        // New issuance per signalled token during time steps
+        // New issuance of tokens during time steps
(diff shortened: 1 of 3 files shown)


abarmat:
Rebased

PR #455, opened 2021-03-19
Dave/update deploy scripts

davekaj:
- Allows all 4 types of deploying to be done from the CLI
- Adds option to build the TX rather than send the tx. This is so it can be used in gnosis multisig, when governor is a multisig
- also updates an event on Governance
- Some small cleaning up of the script


davekaj:
I'm actually just going to merge this into the governance PR, no need to have too separate PRs. The commits are already split nicely. If you want clarity in reading the specific changes that are just deploy scripts, you can always check out the merged PR

PR #457, opened 2021-03-30
Delegation parameters initialization when used from stakeTo()

abarmat:
### Description

The Staking contract has a function that allow indexers to set delegation parameters. Under certain circumstances this function will initialize the wrong address.

### Problem

The function `setDelegationParameters()` in the Staking contract can be used to set the delegation parameters for when rewards are distributed. The Staking contract uses that same function to initialize the delegation parameters when an indexer stakes for the same time to 100% indexer cut.

Initialization can happen within the execution of two public functions: `stake()` and `stakeTo()`. The first will only be called by the indexer, while the second can also be called by third parties addresses depositing stake to the indexer.

The issue lies in that setDelegationParameters() always use msg.sender, this means that when using `stakeTo()` to initialize a delegation pool gives rise to the two-fold incorrect behavior:

- The delegation parameters for indexer remain unset, while we expected them to be set.
- The delegation parameters for msg.sender are unintentionally set.

### Solution

The solution is to have an internal `_setDelegationParameters()` that accepts the indexer address for which we are setting the configuration and use it from `stake()`, `stakeTo()`, and the public `setDelegationParameters()` functions.


trader-payne (commented):
Line 196: wrong quotation, instead of `"text"` you did `""text` . Not sure if that breaks anything though, just wanted to help 🙂

abarmat:
> In addresses.json, Line 196: wrong quotation, instead of `"text"` you did `""text` . Not sure if that breaks anything though, just wanted to help 🙂

Good observation, I fixed it in a different commit on master and needs to rebase now

PR #458, opened 2021-04-01
Disputes with separate slashing percentages for queries and indexing

abarmat:
### Description

The DisputeManager is the contract where actors in the protocol can create disputes about indexers misbehaving. There are two types of disputes: indexing and query ones. Indexing disputes are created based on one allocation that was closed with a wrong POI (proof of indexing) while query disputes are created presenting an attestation for a wrong query returned. After a dispute is created, the Arbitrator can resolve it and slash a percentage of the indexer's stake.

The PR adds a different slashing percentage for Indexing Disputes and Query Disputes.

### Motivation

The current contract uses the same slashing percentage for indexing and query disputes. However, more flexibility is desirable as one indexing offense can carry much more weight than a single bad query returned. Having a too high slashing percentage for a single wrong query can be a disincentive for indexers to serve queries.

### Solution

- Add a type to the dispute to categorize it when someone creates it.
- Create new storage var for a separate indexer and query slashing percentages.
- Add setters for each new slashing governance variable.
- Pick the right slashing percentage when the dispute is resolved.
- Additional minor refactors.

### DIFF
--- a/contracts/disputes/DisputeManager.sol
+++ b/contracts/disputes/DisputeManager.sol
@@ -14,7 +14,7 @@ import "./IDisputeManager.sol";
 
 /*
  * @title DisputeManager
- * @dev Provides a way to align the incentives of participants by having slashing as deterrent
+ * @notice Provides a way to align the incentives of participants by having slashing as deterrent
  * for incorrect behaviour.
  *
  * There are two types of disputes that can be created: Query disputes and Indexing disputes.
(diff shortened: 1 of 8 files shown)


abarmat:
Audited tag + fixes: https://github.com/graphprotocol/contracts/releases/tag/audit%2Fpr-458.
Audited commit: https://github.com/graphprotocol/contracts/commit/b61a9b27eb18eab5ec6bb612244d58c33f2321f2

PR #462, opened 2021-05-04
A contract to exchange vouchers for tokens related to allocations

abarmat:
A contract that works as an exchange where indexers can redeem vouchers signed by an authority to collect tokens for an allocation.


contracts/statechannels/AllocationExchange.sol
```diff
@@ -0,0 +1,103 @@
...
+ * @dev This contract holds tokens that anyone with a voucher signed by the
+ * authority can redeem. The contract validates if the voucher presented is valid
+ * and then sends tokens to the Staking contract by calling the collect() function
+ * passing the voucher allocationID. The contract enforces that only one voucher for
+ * an allocationID can be redeemed.
+ * Only governance can change the authority.
+ */
+contract AllocationExchange is Governed {
+    // An allocation voucher represents a signed message that allows
+    // to redeem an amount of funds from this contract and collect
```
That3Percent, line 23: Minor grammar: "allows to redeem" -> "allows redeeming"

contracts/statechannels/AllocationExchange.sol
```diff
@@ -0,0 +1,103 @@
...
+        AllocationVoucher memory voucher = abi.decode(_msg, (AllocationVoucher));
+
+        // Already collected check
+        require(
+            !allocationsRedeemed[voucher.allocationId],
+            "Collect: allocation already collected"
+        );
+
+        // Signature check
+        bytes32 collectHash = keccak256(abi.encodePacked(voucher.amount, voucher.allocationId));
```
Jannis, line 89: Is the `amount` and `allocationId` order here correct? They are defined the other way around in `AllocationVoucher`.

That3Percent (approved):

contracts/statechannels/AllocationExchange.sol
abarmat: Yes, let's do (allocationID,amount)

contracts/statechannels/AllocationExchange.sol
abarmat: +1

contracts/statechannels/AllocationExchange.sol
```diff
@@ -75,29 +77,26 @@ contract AllocationExchange is Governed {
...
         );
 
         // Signature check
-        bytes32 collectHash = keccak256(abi.encodePacked(voucher.amount, voucher.allocationId));
-        require(
-            authority == ECDSA.recover(collectHash, voucher.signature),
-            "Collect: invalid signer"
-        );
+        bytes32 messageHash = keccak256(abi.encodePacked(_voucher.allocationID, _voucher.amount));
+        bytes32 digest = ECDSA.toEthSignedMessageHash(messageHash);
```
That3Percent, line 131: We don't need this `toEthSignedMessageHash`. As with some of our other designs, this will use a dedicated key that serves no other purpose and holds no funds so there are no collisions. The performance is better this way, and it requires fewer libraries from the signing side. I'm not a big fan of just using `toEthSignedMessageHash` in general, because by itself it only fixes a limited set of collisions. It's an important subset of collisions, but can be a false sense of security without something like EIP-712.

That3Percent (changes requested):

contracts/statechannels/AllocationExchange.sol
abarmat: +1

abarmat: @That3Percent I removed the prefix from the signature, just a plain one now

test/payments/allocationExchange.test.ts
```diff
@@ -0,0 +1,252 @@
...
+    allocationID: string,
+    amount: BigNumber,
+    account: Account,
+  ): Promise<Voucher> {
+    const messageHash = solidityKeccak256(['address', 'uint256'], [allocationID, amount])
+    const messageHashBytes = arrayify(messageHash)
+    const signature = await account.signer.signMessage(messageHashBytes)
+    return {
+      allocationID,
+      amount: amount,
```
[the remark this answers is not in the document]
abarmat: Done

davekaj (approved):

PR #463, opened 2021-05-06
Governance Audit  - Fixes from the open zeppelin audit

davekaj:
Six issues, six commits, six fixes


davekaj:
Closing this, as it was only used to show Zep' how we fixed their requests

PR #469, opened 2021-05-21
Improve configuration and tasks

abarmat:
- Move all tasks to a folder.
- Improve how we load contracts into the environment.
- Fix most of lint error/warnings.
- Improve hardhat config.

### DIFF
--- a/test/upgrade/security.test.ts
+++ b/test/upgrade/security.test.ts
@@ -1,51 +1,51 @@
-import { expect } from 'chai'
-import hre, { upgrades } from 'hardhat'
-import '@nomiclabs/hardhat-ethers'
+// import { expect } from 'chai'
+// import hre, { upgrades } from 'hardhat'
+// import '@nomiclabs/hardhat-ethers'
 
-const { ethers } = hre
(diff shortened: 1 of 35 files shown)


test/upgrade/security.test.ts
```diff
@@ -1,51 +1,51 @@
-import { expect } from 'chai'
-import hre, { upgrades } from 'hardhat'
-import '@nomiclabs/hardhat-ethers'
+// import { expect } from 'chai'
+// import hre, { upgrades } from 'hardhat'
```
davekaj, line 2: Why did this file get commented out?

davekaj (approved):
just wondering about the commented out file, otherwise good to go

test/upgrade/security.test.ts
abarmat: The test helped with testing the first Staking contract upgrade we did. I'm planning to do something different that will help with "generically" test any upgrade. Thinking about creating a docker image running hardhat node, deploy the last deployed version of contracts there, then checkout the new version and upgrade to see if everything works.

PR #474, opened 2021-06-01
Stable cut for indexing and query rewards

abarmat:
### Summary

New delegations and undelegations dramatically change the staking income for indexers and validators unless the cut rate is regularly adjusted.

Why is this happening? Staking income for the indexer and delegators is pooled before the Indexer cut is calculated.

This PR is inspired by https://forum.thegraph.com/t/proposal-to-change-the-indexer-cut-mechanism/1422

### Solution

Distribute any new rewards based on the delegation to total-stake ratio at the moment the allocation was created. In addition, the indexer cut will be defined as the share of delegator rewards the indexer gets.

**Example:**
If you delegate to an indexer with a ‘rewards cut’ and ‘fee cut’ of 10%, it means you will get 90% of whatever your staked GRT earns and your indexer will get the other 10%. If you run an indexer, you will earn your pro-rata staking income in full and a 10% commission on your delegators’ staking income.

### Features

- Save the delegators rewards cut at the moment when the allocation is created. This avoids any manipulation of the cuts before closing an allocation. This means changing the allocation struct.
- Calculate the`indexerDelegationRatio` as the ratio of delegation to total stake for a particular indexer.
- Rename some delegation pool variables to make them clearer.
- Improve comments.
- Use a single function to collect any delegators rewards.

### DIFF
--- a/contracts/staking/IStaking.sol
+++ b/contracts/staking/IStaking.sol
@@ -38,8 +38,8 @@ interface IStaking is IStakingData {
     function setDelegationRatio(uint32 _delegationRatio) external;
 
     function setDelegationParameters(
-        uint32 _indexingRewardCut,
-        uint32 _queryFeeCut,
+        uint32 _indexRewardsCut,
+        uint32 _queryRewardsCut,
         uint32 _cooldownBlocks
     ) external;
 
--- a/contracts/staking/Staking.sol
+++ b/contracts/staking/Staking.sol
@@ -1168,7 +1168,6 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
         pure
         returns (uint32)
     {
-        if (_indexerDelegationRatio == 0 || _indexerRewardsCut == MathUtils.MAX_PPM) return 0;
         return
             MathUtils.percentOfPercent(
                 MathUtils.percentFlip(_indexerRewardsCut),
--- a/contracts/staking/Staking.sol
+++ b/contracts/staking/Staking.sol
@@ -22,9 +22,6 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
     using Stakes for Stakes.Indexer;
     using Rebates for Rebates.Pool;
 
-    // 100% in parts per million
-    uint32 private constant MAX_PPM = 1000000;
-
     // -- Events --
 
     /**
@@ -284,7 +281,7 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
      */
     function _setCurationPercentage(uint32 _percentage) private {
         // Must be within 0% to 100% (inclusive)
-        require(_percentage <= MAX_PPM, ">percentage");
+        require(_percentage <= MathUtils.MAX_PPM, ">percentage");
         curationPercentage = _percentage;
         emit ParameterUpdated("curationPercentage");
     }
@@ -303,7 +300,7 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
      */
     function _setProtocolPercentage(uint32 _percentage) private {
         // Must be within 0% to 100% (inclusive)
-        require(_percentage <= MAX_PPM, ">percentage");
+        require(_percentage <= MathUtils.MAX_PPM, ">percentage");
         protocolPercentage = _percentage;
         emit ParameterUpdated("protocolPercentage");
     }
@@ -417,8 +414,8 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
         uint32 _cooldownBlocks
     ) private {
         // Incentives must be within bounds
-        require(_queryFeeCut <= MAX_PPM, ">queryFeeCut");
-        require(_indexingRewardCut <= MAX_PPM, ">indexingRewardCut");
+        require(_queryFeeCut <= MathUtils.MAX_PPM, ">queryFeeCut");
+        require(_indexingRewardCut <= MathUtils.MAX_PPM, ">indexingRewardCut");
 
         // Cooldown period set by indexer cannot be below protocol global setting
         require(_cooldownBlocks >= delegationParametersCooldown, "<cooldown");
@@ -498,7 +495,7 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
      */
     function _setDelegationTaxPercentage(uint32 _percentage) private {
         // Must be within 0% to 100% (inclusive)
-        require(_percentage <= MAX_PPM, ">percentage");
+        require(_percentage <= MathUtils.MAX_PPM, ">percentage");
         delegationTaxPercentage = _percentage;
         emit ParameterUpdated("delegationTaxPercentage");
     }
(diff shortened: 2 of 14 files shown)


test/lib/testHelpers.ts
```diff
@@ -8,14 +8,38 @@ import { EpochManager } from '../../build/types/EpochManager'
...
+
 export const randomHexBytes = (n = 32): string => hexlify(randomBytes(n))
 export const randomAddress = (): string => getAddress(randomHexBytes(20))
 
+// Core Math
+
+export const MAX_PPM = toBN('1000000')
+export const totalRatio = (a: BigNumber, b: BigNumber): BigNumber => a.mul(MAX_PPM).div(a.add(b))
+export const percentageFlip = (ppm: BigNumber): BigNumber => MAX_PPM.sub(ppm)
+export const percentageOf = (ppm: BigNumber, value: BigNumber): BigNumber =>
```
davekaj, line 29: seems this is not 100% equal to the solidity implementation, as it could work with numbers above 100%

contracts/staking/Staking.sol
```diff
@@ -22,8 +22,9 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
     using Stakes for Stakes.Indexer;
     using Rebates for Rebates.Pool;
 
-    // 100% in parts per million
-    uint32 private constant MAX_PPM = 1000000;
+    // For safety purposes we define an activation epoch for using the new rewards
+    // distribution formula
+    uint256 private constant GIP_ACTIVATION_EPOCH = 0; // TODO: change before deployment of implementation
 
     // -- Events --
```
davekaj, line 30: See a few lines down, `event DelegationParametersUpdated` - let's update the names of the event params `queryFeeCut` and `indexingRewardCut` to the new ones

test/utils/math-utils.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    const percentTests: BigNumber[] = [
+      0,
+      1,
+      2,
+      1e6 * 0.1,
+      1e6 * 0.25,
+      1e6 * 0.5,
+      1e6 * 0.75,
+      1e6 * 0.9,
+      1e6 * 1.0,
```
davekaj, line 88: testing `1e6 * 1.1` here would have created an inconsistency between the two functions, i believe

davekaj (changes requested):
Some simple changes requested

contracts/staking/libs/MathUtils.sol
```diff
@@ -42,4 +45,49 @@ library MathUtils {
...
+     */
+    function percentMul(uint32 a, uint32 b) internal pure returns (uint32) {
+        return toPercent(percentOf(a, uint256(b)));
+    }
+
+    /**
+     * @dev Returns (1 - percentage) in parts-per-million.
+     * @param percentage Percentage (PPM)
+     */
+    function percentFlip(uint32 percentage) internal pure returns (uint32) {
```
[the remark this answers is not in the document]
abarmat: Makes sense

contracts/staking/libs/MathUtils.sol
```diff
@@ -42,4 +45,49 @@ library MathUtils {
...
+        require(value <= MAX_PPM, "PercentCast: out of bounds");
+        return uint32(value);
+    }
+
+    /**
+     * @dev Returns the value after applying percentage with parts-per-million precision.
+     * @param percentage Percentage (PPM)
+     * @param value Value to calculate the percentage of
+     */
+    function percentOf(uint32 percentage, uint256 value) internal pure returns (uint256) {
```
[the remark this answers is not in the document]
abarmat: Good idea

contracts/staking/libs/MathUtils.sol
```diff
@@ -42,4 +45,49 @@ library MathUtils {
...
+     */
+    function toPercent(uint256 value) internal pure returns (uint32) {
+        require(value <= MAX_PPM, "PercentCast: out of bounds");
+        return uint32(value);
+    }
+
+    /**
+     * @dev Returns the value after applying percentage with parts-per-million precision.
+     * @param percentage Percentage (PPM)
+     * @param value Value to calculate the percentage of
```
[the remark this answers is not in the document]
abarmat: +1

contracts/staking/Staking.sol
```diff
@@ -22,8 +22,9 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
     using Stakes for Stakes.Indexer;
     using Rebates for Rebates.Pool;
 
-    // 100% in parts per million
-    uint32 private constant MAX_PPM = 1000000;
+    // For safety purposes we define an activation epoch for using the new rewards
+    // distribution formula
+    uint256 private constant GIP_ACTIVATION_EPOCH = 0; // TODO: change before deployment of implementation
```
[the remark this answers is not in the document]
abarmat: I wasn't sure yet about the GIP number and the epoch, so I left it with a TODO to remember changing it

contracts/staking/Staking.sol
```diff
@@ -1147,6 +1164,32 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
         );
     }
 
+    /**
+     * @dev Calculate the delegator rewards cut from the total staked amount (indexer + delegations).
+     *
+     * Formula:
+     * 1) indexerDelegationTotalRatio = delegatedStake / totalStake
```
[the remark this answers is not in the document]
abarmat: I agree

contracts/staking/Staking.sol
abarmat: Can we break the subgraph by doing that?

test/utils/math-utils.test.ts
abarmat: I'll force the error and then fix the function in the Typescript code

contracts/staking/Staking.sol
davekaj: It would. we can handle it in the subgraph ahead of time though by changing:
```
    indexer.indexingRewardCut = event.params.indexingRewardCut.toI32()
    indexer.queryFeeCut = event.params.queryFeeCut.toI32()
```

to 
```
    indexer.indexingRewardCut = event.parameters[1] as i32
    indexer.queryFeeCut = event.parameters[2] as i32
    ```

Comment (davekaj) [reply contracts/staking/Staking.sol:30]: we could also use `ethereum.decode()` instead of `parameters` https://github.com/graphprotocol/graph-ts/pull/172/files

File: contracts/staking/Staking.sol [hunk 9]
```diff
@@ -1558,36 +1592,37 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
      * @param _subgraphDeploymentID Subgraph deployment updated
      */
     function _updateRewards(bytes32 _subgraphDeploymentID) private returns (uint256) {
-        IRewardsManager rewardsManager = rewardsManager();
-        if (address(rewardsManager) == address(0)) {
```
davekaj, line 1562: what is the reason for removing the 0 check?

contracts/staking/Staking.sol
abarmat: I think it's a good moment to discuss this as I can change it. We can use this opportunity to make it clearer or just leave it as it is.

The idea of using `queryRewards` instead of `queryFee` is:
- queryRewards: Tokens that comes out of the rebate pool when the allocation is claimed.
- queryFees: Total tokens collected on the allocation.

These two things not necessarily will match, depends on the Cobbs-Douglas formula.

contracts/staking/IStaking.sol
```diff
@@ -38,8 +38,8 @@ interface IStaking is IStakingData {
     function setDelegationRatio(uint32 _delegationRatio) external;
 
     function setDelegationParameters(
-        uint32 _indexingRewardCut,
-        uint32 _queryFeeCut,
+        uint32 _indexRewardsCut,
+        uint32 _queryRewardsCut,
```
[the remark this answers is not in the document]
abarmat: See https://github.com/graphprotocol/contracts/pull/474/files#r674383832

pcarranzav:

PR #475, opened 2021-06-07
Add multicall to staking contract

abarmat:
### Summary

Expose a way to batch multiple calls into a single transaction. It provides great flexibility for indexer agents to combine multiple functions in different ways. It also reduce the gas cost by saving the initial gas, and in some cases, accessing a "used" slot by the other bundled transactions.

### Solution

By using a multicall a user can batch an arbitrary number of operations into a single transaction.

### Breaking

This implementation removes the following functions from the interface to save contract bytecode. They can be now used through a multicall.

- closeAllocationMany()
- closeAndAllocate()
- claimMany()
- allocate() `since allocateFrom() already covers this use case and it's not currently used in production`

### DIFF
--- a/contracts/staking/IStaking.sol
+++ b/contracts/staking/IStaking.sol
@@ -113,24 +113,9 @@ interface IStaking is IStakingData {
 
     function closeAllocation(address _allocationID, bytes32 _poi) external;
 
-    function closeAllocationMany(CloseAllocationRequest[] calldata _requests) external;
-
-    function closeAndAllocate(
-        address _oldAllocationID,
-        bytes32 _poi,
-        address _indexer,
-        bytes32 _subgraphDeploymentID,
-        uint256 _tokens,
-        address _allocationID,
-        bytes32 _metadata,
-        bytes calldata _proof
-    ) external;
-
     function collect(uint256 _tokens, address _allocationID) external;
 
     function claim(address _allocationID, bool _restake) external;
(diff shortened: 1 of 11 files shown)


contracts/staking/IStaking.sol
```diff
@@ -107,25 +113,10 @@ interface IStaking is IStakingData {
 
     function closeAllocation(address _allocationID, bytes32 _poi) external;
 
-    function closeAllocationMany(CloseAllocationRequest[] calldata _requests) external;
```
davekaj, line 110: whats the significance on us removing these functions on the Indexer Agent? Do you have to write custom code in the agent to use mutlicall for :

- closeAllocationMany()
- closeAndAllocate()
- claimMany()

davekaj (approved):
Just a question

contracts/staking/IStaking.sol
abarmat: The agent does not use any of those functions yet. Anyone calling those is because they are using something custom. Whoever is using that interface will need to use multicall, any upgrade will need to be communicated with time.

davekaj: okay that is good for me!

maybe we could keep the functions if it were a smaller contract. but staking is so big - it will be useful to delete these for future upgrades

abarmat:
Rebased

PR #485, opened 2021-07-19
Allow batch transactions using multicall in the GNS

abarmat:
### Summary

By using a multicall a user can batch multiple operations into a single transaction. 
An example of this is if the owner publishing a subgraph decides to mint some signal on that subgraph.


test/gns.test.ts
```diff
@@ -983,4 +983,68 @@ describe('GNS', () => {
       await gns.connect(me.signer).mintNSignal(me.address, 1, toGRT('10'), 0)
     })
   })
+
+  describe('batch calls', function () {
```
davekaj, line 987: looks good. i would like to add a few tests to play it safe. i want to add the following sequences:

- publishSubgraph, publishSubgraph
- publishSubgraph, publishNewVersion
- publishSubgraph, mintNSignal, burnNsignal

I think it will be fine. I can write the tests for these

contracts/discovery/GNS.sol
```diff
@@ -136,16 +142,28 @@ contract GNS is GNSV1Storage, GraphUpgradeable, IGNS {
...
+
     /**
-    @dev Modifier that allows a function to be called by owner of a graph account
-    @param _graphAccount Address of the graph account
-    */
-    modifier onlyGraphAccountOwner(address _graphAccount) {
+     * @dev Check if the owner is the graph account
+     * @param _graphAccount Address of the graph account
+     */
+    function _isGraphAccountOwner(address _graphAccount) private view {
```
davekaj, line 151: does this change save on gas?

davekaj (changes requested):
I am only requesting changes as I want to add some tests. I can add the tests. they are just an extra percaution I want to add

contracts/discovery/GNS.sol
abarmat: Yes!

test/gns.test.ts
abarmat: Good idea

abarmat:
Rebased

nenadjaja:
This is pretty cool @abarmat ! It will be easy to implement it on our end :-D

PR #487, opened 2021-07-20
Allow unstake passing a larger amount than available

abarmat:
### Description

When an indexer decides to unstake their stake completely there might be an opportunity to frontrun that operation through the `stakeTo()` function, making unstake to fail with stake under minimum tokens.

### Solution

Allow `unstake()` to accept a larger amount than the currently staked tokens. The function will try to unstake the passed amount capped to the available stake.


davekaj (approved):
Looks good to me. Only thing I saw was the same thing Heiko mentioned in slack

abarmat:
Audited tag + fixes: https://github.com/graphprotocol/contracts/releases/tag/audit/pr-487
Audited commit: https://github.com/graphprotocol/contracts/pull/487/commits/b2a4c171e3153b244ec2918480693ffaffefb59e

PR #492, opened 2021-08-10
Average delegation unbonding on multiple requests

abarmat:
### Summary

When a delegator undelegates some tokens a `delegationUnbondingPeriod` is used to calculate the epoch when the tokens will be unlocked. On each new undelegation the period is reset to the unbonding period even if there are tokens waiting to be unlocked. This PR change this behaviour to use a weighted average of time and tokens like on the indexer thawing period.

### Current Behavior

It currently works by resetting the unlock epoch on each undelegation.

Example:
- Epoch 1: Undelegation for 100 GRT, unlock epoch is set to 29
- Epoch 2: 27 epochs to go for unlocking the tokens
- Epoch 3: 26 epochs to go for unlocking the tokens 
- Epoch 4: Undelegation for 50 GRT, unlock epoch is set to 32 <- the period is reset for the 150 GRT total tokens

### Solution

Use a weighted average formula that consider the time some tokens have already been locked, weighted by the old tokens and the new tokens being undelegated. The formula will round up any fractional epoch value as there is no such thing as half an epoch.

### DIFF
--- a/contracts/staking/Staking.sol
+++ b/contracts/staking/Staking.sol
@@ -1324,6 +1324,34 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
     }
 
     /**
+     * @dev Get the epoch when tokens will be unlocked based on a new undelegation.
+     * @param _del Delegation
+     * @param _tokens Amount of tokens to undelegate
+     * @return Epoch when the undelegated tokens unlock
+     */
+    function _getDelegatedTokensLockedUntil(Delegation memory _del, uint256 _tokens)
+        internal
+        view
+        returns (uint256)
+    {
+        uint256 lockingPeriod = delegationUnbondingPeriod;
+        uint256 currentEpoch = epochManager().currentEpoch();
+        if (_del.tokensLocked > 0) {
+            // Calculate weighted average multiplying epochs by 10 to add extra precision
+            // and be able to round the final number up
+            lockingPeriod = MathUtils.weightedAverage(
+                MathUtils.diffOrZero(_del.tokensLockedUntil, currentEpoch).mul(10), // Remaining thawing period
+                _del.tokensLocked, // Weighted by remaining unstaked tokens
+                uint256(lockingPeriod).mul(10), // Thawing period
+                _tokens // Weighted by new tokens to unstake
+            );
+            // Round up the result
+            lockingPeriod = (lockingPeriod % 10 > 0) ? lockingPeriod / 10 + 1 : lockingPeriod / 10;
+        }
+        return currentEpoch.add(lockingPeriod);
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


contracts/staking/Staking.sol
```diff
@@ -1323,6 +1323,34 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
...
+        view
+        returns (uint256)
+    {
+        uint256 lockingPeriod = delegationUnbondingPeriod;
+        uint256 currentEpoch = epochManager().currentEpoch();
+        if (_del.tokensLocked > 0) {
+            // Calculate weighted average multiplying epochs by 10 to add extra precision
+            // and be able to round the final number up
+            lockingPeriod = MathUtils.weightedAverage(
+                MathUtils.diffOrZero(_del.tokensLockedUntil, currentEpoch).mul(10), // Remaining thawing period
```
davekaj, line 1343: One thing I am thinking is to make the value 10 as a variable, like `uint256 extraPrecision = 10` 

It is something an auditor mentioned once before, and I liked it, it makes the code more readable. 

I feel it is not worth it to make it a global variable. It is not important enough. Maybe just a local function variable. Wdyt?

test/lib/testHelpers.ts
```diff
@@ -87,6 +89,24 @@ export const advanceToNextEpoch = async (epochManager: EpochManager): Promise<vo
...
 
+// Math
+
+export const MAX_PPM = toBN('1000000')
+
+export function weightedAverage(
+  valueA: BigNumber,
+  valueB: BigNumber,
+  weightA: BigNumber,
+  weightB: BigNumber,
```
davekaj, line 100: It might help to make this function exactly the same params as the solidity function:

```
    function weightedAverage(
        uint256 valueA,
        uint256 weightA,
        uint256 valueB,
        uint256 weightB
    ) 
```
    

Could prevent future bugs

test/staking/delegation.test.ts
```diff
@@ -446,6 +464,7 @@ describe('Staking::Delegation', () => {
       })
 
       it('should undelegate properly when multiple delegations', async function () {
+        await staking.setDelegationUnbondingPeriod(28)
```
davekaj, line 467: Was this missing from before?

davekaj (approved):
Looks good, I approved, but I still have a few small questions

test/lib/testHelpers.ts
abarmat: Good idea, I'll reorder the input values in the JS function

contracts/staking/Staking.sol
abarmat: I'll do it

pcarranzav:

PR #505, opened 2021-09-23
Use a minimal proxy for the curation shares ERC20

abarmat:
### Description

Signal on a subgraph deployment is represented by a subgraph-specific ERC20 token. When signal is minted for a subgraph deployment for the first time, the Curation contract deploys a new ERC20 token. This is costly, about 1.2M gas.

### Solution

Use a Minimal Proxy to clone the Signal-ERC20 based on an implementation. First mint goes from  1,230,118 gas to 432,840


kajpust:
Great job

PR #506, opened 2021-09-23
Validate the POI is from the canonical chain and close to chain head

abarmat:


contracts/staking/Staking.sol
```diff
@@ -20,13 +21,15 @@ import "./libs/Stakes.sol";
...
-contract Staking is StakingV2Storage, GraphUpgradeable, IStaking {
+contract Staking is StakingV2Storage, GraphUpgradeable, IStaking, Multicall {
     using SafeMath for uint256;
     using Stakes for Stakes.Indexer;
     using Rebates for Rebates.Pool;
 
     // 100% in parts per million
     uint32 private constant MAX_PPM = 1000000;
+    // Range to use for valid POI blocks from latest block number
+    uint256 private constant POI_VALID_BLOCK_RANGE = 256;
```
azf20, line 32: This is only ~ an hour - given that the PoI needs to be for the first block in an epoch, does this mean that indexers will only be able to close allocations in the first hour of the epoch (or it won't be able to find the blockhash)?

azf20:
Is 256 blocks enough time, or does it create a very narrow window when it is possible to close allocations (the first 256 blocks of the epoch)?
Also worried that this on-chain validation here won't extend to indexing other chains, which we want to support soon.

abarmat:
@azf20 the max blocks are 256 which is roughly one hour, and about 50 min post merge. For the context of data freshness I think it is enough, while it can be challenging to increase if we ever need to.

The on-chain validation is very Ethereum specific, if we were to validate through a contract that some blockhash presented is from the canonical chain in the context of multiblochain we would need to do something else, like keeping track of that in a different contract and use that for validation. This is probably the main point against enforcing through the blockhash.

azf20:
Sure, I guess my thinking was as follows:
- Close allocation needs to use the first block of the epoch (first block of a 24 hour period)
- With this new validation, the block needs to be within the last ~hour on chain
- This means that indexers will only be able to close allocations for one hour per day (start of the epoch plus one hour)
This seems like quite a big constraint (vs. the current state where you can close an allocation at any time during the epoch)

contracts/staking/Staking.sol
abarmat: The Arbitration Charter needs to change to allow any indexer to be able to provide a proof within the N amount of blocks from the block they sent the transaction instead of using the start block of epoch.

abarmat:
The idea is to change the Charter so instead of using the first block of the current epoch you just need to present a fresh POI, that meaning one within the bounds of the 256 blocks.

azf20:
OK got it, that makes sense. Sorry I maybe landed in this PR without full context!

pcarranzav:
This would require a lot of changes to fix conflicts and to work for multiple chains with the EBO, so closing for now

PR #508, opened 2021-09-28
Initialization phase for a Subgraph

jona:


graph.config.yml
```diff
@@ -48,6 +48,8 @@ contracts:
       reserveRatio: 500000 # 50% (parts per million)
       curationTaxPercentage: 25000 # 2.5% (parts per million)
       minimumCurationDeposit: "1000000000000000000" # 1 GRT
+      initializationPeriod: 432000 # Seconds = 5 Days
+      initializationExitPeriod: 2592000 # Seconds = 30 Days
```
jona, line 52: These will need to be adjusted once we finalize the GIP. They will also be affected by the upcoming Capital Gains Tax GIP as they both work in harmony to balance out incentives/disincentives.

contracts/curation/Curation.sol
```diff
@@ -463,4 +530,28 @@ contract Curation is CurationV1Storage, GraphUpgradeable, ICuration {
...
+     */
+    function _getEffectiveReserveRatio(uint256 _createdAt) private view returns (uint32) {
+        uint32 effectiveReserveRatio = defaultReserveRatio;
+
+        if (block.timestamp <= _createdAt.add(initializationPeriod)) {
+            effectiveReserveRatio = MAX_PPM;
+        } else if (
+            block.timestamp <= (_createdAt.add(initializationPeriod).add(initializationExitPeriod))
+        ) {
+            uint256 blockDiff = block.timestamp.sub(_createdAt.add(initializationPeriod));
```
abarmat, line 547: Just a warning, this is a `timeDiff` now instead of blocks

contracts/curation/Curation.sol
```diff
@@ -395,6 +456,8 @@ contract Curation is CurationV1Storage, GraphUpgradeable, ICuration {
         // Get curation pool tokens and signal
         CurationPool memory curationPool = pools[_subgraphDeploymentID];
 
+        uint32 _effectiveReserveRatio = _getEffectiveReserveRatio(curationPool.createdAt);
```
abarmat, line 459: Something that happens with this implementation is that it breaks the possibility to mint at the Curation contract level. 

If the Curation.mint() function is used directly by any user the `createdAt` variable will be zero. I think we should store `createdAt` on the subgraph struct whenever the first mint happens as part of initialization.

contracts/curation/Curation.sol
```diff
@@ -463,4 +530,28 @@ contract Curation is CurationV1Storage, GraphUpgradeable, ICuration {
...
+
+    /**
+     * @dev Calculate reserve ratio based on initialization phase
+     * @param _createdAt When NameCurationPool was created
+     * @return Reserve ratio
+     */
+    function _getEffectiveReserveRatio(uint256 _createdAt) private view returns (uint32) {
+        uint32 effectiveReserveRatio = defaultReserveRatio;
+
+        if (block.timestamp <= _createdAt.add(initializationPeriod)) {
```
abarmat, line 542: We could store this intermediate calculation in a variable as it is reused in different places `_createdAt.add(initializationPeriod)`

contracts/curation/Curation.sol
```diff
@@ -463,4 +530,28 @@ contract Curation is CurationV1Storage, GraphUpgradeable, ICuration {
...
         }
     }
+
+    /**
+     * @dev Calculate reserve ratio based on initialization phase
+     * @param _createdAt When NameCurationPool was created
+     * @return Reserve ratio
+     */
+    function _getEffectiveReserveRatio(uint256 _createdAt) private view returns (uint32) {
+        uint32 effectiveReserveRatio = defaultReserveRatio;
```
abarmat, line 540: One issue I see with reading directly from `defaultReserveRatio` is that if Governance changes the reserve ratio at any point all the calculations from previously initialized subgraphs will change. 

The current behaviour is to store the reserve ratio to use in `Subgraph.reserveRatio` whenever it is initialized. I believe we should still do that during the first mint and use it from here.

What do you think?

contracts/discovery/GNS.sol
```diff
@@ -286,6 +289,10 @@ contract GNS is GNSV1Storage, GraphUpgradeable, IGNS, Multicall {
             "GNS: Cannot publish a new version with the same subgraph deployment ID"
         );
 
+        NameCurationPool storage namePool = nameSignals[_graphAccount][_subgraphNumber];
+
+        curation().setCreatedAt(_subgraphDeploymentID, namePool.createdAt);
```
abarmat, line 294: I don't feel comfortable with this GNS->Curation interaction that makes Curation completely dependent on the GNS, but I'm trying to think if there is a better option and couldn't think of anything better yet.

contracts/curation/Curation.sol
abarmat: When we upgrade the Curation contract with this new implementation, all the existing subgraphs deployments will have createdAt set at 0. Maybe we should add if-clause to return the plain reserve ratio when Subgraph.createdAt = 0

PR #515, opened 2021-10-13
Updateable reserveRatio for GNS

jona:
## Summary

We currently have no way to set `reserveRatio` in the GNS contract through governance.

## Solution

Add a `setReserveRatio` function that allows us to set `reserverRatio`

### DIFF
--- a/contracts/discovery/IGNS.sol
+++ b/contracts/discovery/IGNS.sol
@@ -21,7 +21,7 @@ interface IGNS {
 
     function setOwnerTaxPercentage(uint32 _ownerTaxPercentage) external;
 
-    function setReserveRatio(uint32 _defaultReserveRatio) external;
+    function setReserveRatio(uint32 _reserveRatio) external;
 
     // -- Publishing --
 


test/gns.test.ts
```diff
@@ -708,6 +712,26 @@ describe('GNS', () => {
...
+          .connect(me.signer)
+          .publishNewVersion(
+            me.address,
+            0,
+            subgraph1.subgraphDeploymentID,
+            subgraph1.versionMetadata,
+          )
+
+        const pool = await gns.nameSignals(me.address, 0)
+        const reserveRatio = pool[3]
```
abarmat, line 731: `gns.nameSignals(me.address, 0)` returns an object that lets you access attributes by both position and key:

```
[
  BigNumber { _hex: '0x858ffc3f497c9112', _isBigNumber: true },
  BigNumber { _hex: '0x874391d0dd1fdf42', _isBigNumber: true },
  '0x71a94176484d9e0d2ff853a92e162a0493fafa1823a0841400c054769bd895a1',
  10,
  false,
  BigNumber { _hex: '0x00', _isBigNumber: true },
  vSignal: BigNumber { _hex: '0x858ffc3f497c9112', _isBigNumber: true },
  nSignal: BigNumber { _hex: '0x874391d0dd1fdf42', _isBigNumber: true },
  subgraphDeploymentID: '0x71a94176484d9e0d2ff853a92e162a0493fafa1823a0841400c054769bd895a1',
  reserveRatio: 10,
  disabled: false,
  withdrawableGRT: BigNumber { _hex: '0x00', _isBigNumber: true }
]
```

You should be able to do something like:
`expect(pool.reserveRatio).eq(value)`

contracts/discovery/GNS.sol
```diff
@@ -286,6 +311,8 @@ contract GNS is GNSV1Storage, GraphUpgradeable, IGNS, Multicall {
             "GNS: Cannot publish a new version with the same subgraph deployment ID"
         );
 
+        nameSignals[_graphAccount][_subgraphNumber].reserveRatio = reserveRatio;
```
abarmat, line 314: Maybe we could move this to the `_upgradeNameSignal` function, right before `namePool.subgraphDeploymentID = _newSubgraphDeploymentID`. What do you think?

test/gns.test.ts
jona: I was following the pattern used here: https://github.com/graphprotocol/contracts/blob/dev/test/gns.test.ts#L157-L159

Wanted to make sure to edit and mutate as little as possible with this PR

contracts/discovery/GNS.sol
jona: Yeah good idea! We are using `namePool` in that function already :)

test/gns.test.ts
abarmat: I understand. I moved away from using positional arguments on a refactor I did for this PR https://github.com/graphprotocol/contracts/pull/497

We can use the old pattern here and then eventually when audit passes on #497 and if governance decides to move forward with that upgrade we can rebase/merge.

PR #541, opened 2022-02-01
Configure AllocationExchange as assetHolder

Theodus:


abarmat:
Closes https://github.com/graphprotocol/contracts/issues/534

abarmat (approved):

Theodus:
Do you think these CI failures are related? I don't think they are

abarmat:
@Theodus it doesn't seem related, I tried it locally and it works. I'm trying to figure out if it has something to do with a timeout.

PR #543, opened 2022-02-09
Allow anyone to close allocation after max allocation period

abarmat:
### Goal

Provide a mechanism so that anyone can close stale allocations.

### Motivation

Stale allocations affect the stability of the network when an indexer that has a big proportional allocated stake on a subgraph does not keep it active, as other indexer won't want to allocate if it already has a big allocation. Stale allocation also reduce the speed of how value flows through the protocol, reducing the frequency at which curators and delegators get their share of rewards/fees. 

Currently delegators for an indexer are the only one that can close them, but this is an unnecessary validation considering that anyone could turn into a delegator, even with the smallest amount of tokens, and then close it.

### Reference

Forum: https://forum.thegraph.com/t/rewarded-force-close-mechanism-to-eliminate-stale-allocations/2951/8


abarmat:
Closing as it was superseded by this implementation https://github.com/graphprotocol/contracts/pull/549

PR #544, opened 2022-02-11
Present a POI (proof of indexing) when opening an allocation

abarmat:
This PR is built on top of rebased changes from https://github.com/graphprotocol/contracts/pull/475


pcarranzav:

PR #548, opened 2022-03-07
Query versioning for dispute resolution

abarmat:
### Motivation

This change to the Dispute Manager is intended to allow multiple query attestation formats that involve different query versions as described in this GIP https://forum.thegraph.com/t/gip-0024-query-versioning/3145

### Changes

- Add a `bytes32` **domainSeparator** field to the attestation payload submitted on both `createQueryDispute` and `createQueryDisputeConflict`.
- Use the passed **domainSeparator** to build the message hash used for recovering the signer.
- Deprecate the domain separator with fixed versioning from the contract storage.


That3Percent (approved):

That3Percent (commented):
These are still not consistent. Some end with `SIZE_LENGTH` and some with `LENGTH`

That3Percent (dismissed):

pcarranzav:

PR #549, opened 2022-03-14
Create altruistic allocations and close stale allocations

abarmat:
Supersedes: https://github.com/graphprotocol/contracts/pull/543

### Goal

Provide a mechanism so that indexers can opt-out of rewards by allocating zero tokens to a subgraph. In addition, this implementation allows anyone to close stale allocations after the max allocation period, however, nobody can close an altruistic allocation.

### Motivation

1) Stale allocations affect the stability of the network when an indexer that has a big proportional allocated stake on a subgraph does not keep it active, as other indexer won't want to allocate if it already has a big allocation. Stale allocation also reduce the speed of how value flows through the protocol, reducing the frequency at which curators and delegators get their share of rewards/fees. Currently delegators for an indexer are the only one that can close them, but this is an unnecessary validation considering that anyone could turn into a delegator, even with the smallest amount of tokens, and then close it.

> The mechanism for closing state allocations after the `maxAllocationEpoch` was introduced to avoid indexers holding off rewards forever and never distributing them to delegators. Additionally, it promotes liveness in the protocol state changes that effect other stakeholders decisions.

2) Indexers might want to index and serve queries on a particular subgraph altruistically by opting out of rewards. Today they can effectively do that by presenting an empty POI on closeAllocation, however this is not the most efficient way to do it. By allowing them allocating zero tokens, we can skip all the code paths for updating rewards getting a huge gas improvement.

### Highlight

- Indexers can allocate zero-tokens to a subgraph to advertise they are indexing and serving queries but opted out of rewards.
- Even if indexers allocate zero token they are required to have the minimum stake.
- Altruistic allocations cannot be closed after the `maxAllocationEpoch` by the public, only by indexer or operator.
- Anyone can now close a standard allocation after the `maxAllocationEpoch`.

# Gas Efficiency

Altruistic allocations are roughly 50% more gas efficient because the contract doesn't need to recalculate rewards snapshots.

**Standard Allocation**
```
|  Staking  ·  allocate         ·  277077  │
|  Staking  ·  closeAllocation  ·  233146  │
```

**Zero Allocation**
```
|  Staking  ·  allocate         ·  144355  │
|  Staking  ·  closeAllocation  ·  129128  │
```

### Reference

Forum: https://forum.thegraph.com/t/rewarded-force-close-mechanism-to-eliminate-stale-allocations/2951/8


test/staking/allocation.test.ts
```diff
@@ -15,6 +15,7 @@ import {
   toBN,
   toGRT,
   Account,
+  advanceEpochs,
 } from '../lib/testHelpers'
 
 const { AddressZero, HashZero } = constants
```
tmigone, line 21: Minor nit, `HashZero` can be removed here

tmigone, line 21: Minor nit, `HashZero` can be removed here

contracts/staking/libs/Rebates.sol
```diff
@@ -86,7 +87,7 @@ library Rebates {
         uint256 rebateReward = 0;
 
         // Calculate the rebate rewards for the indexer
-        if (pool.fees > 0) {
+        if (pool.fees > 0 && pool.effectiveAllocatedStake > 0) {
```
pcarranzav, line 90: Could this be `pool.exists()`? Or, should the definition of `pool.exists` change? (I noticed now it's possible that a pool went through `pool.init` but its effective allocated state is still zero, so not sure how to interpret it or what's the best way to account for it)

test/lib/testHelpers.ts
```diff
@@ -83,6 +83,12 @@ export const advanceToNextEpoch = async (epochManager: EpochManager): Promise<vo
   return advanceBlocks(epochLen.sub(blocksSinceEpoch))
 }
 
+export const advanceEpochs = async (epochManager: EpochManager, n: number): Promise<void> => {
+  for (let i = 0; i < n + 1; i++) {
```
pcarranzav, line 87: Isn't this advancing `n+1` epochs? i.e. if `n == 1`, this will loop twice.

test/staking/allocation.test.ts
```diff
@@ -484,193 +577,206 @@ describe('Staking:Allocation', () => {
...
-
-    it('reject close before at least one epoch has passed', async function () {
-      const tx = staking.connect(indexer.signer).closeAllocation(allocationID, poi)
-      await expect(tx).revertedWith('<epochs')
-    })
-
-    it('reject close if not the owner of allocation', async function () {
-      // Move at least one epoch to be able to close
-      await advanceToNextEpoch(epochManager)
+    for (const tokensToAllocate of [toBN(100), toBN(0)]) {
```
pcarranzav, line 582: nit: I see how this is practical to make the code shorter, but I wonder if it isn't cleaner to just put the two cases separately? Otherwise below you end up adding checks for whether the allocation is zero tokens or not, and it becomes a bit harder to follow what the test is doing.

test/staking/allocation.test.ts
tmigone: VSCode github messed up the lines, I'm referring to L21

pcarranzav (commented):
Just a few details for now, but otherwise I think it looks good (still taking a closer look)

contracts/staking/Staking.sol
```diff
@@ -1119,8 +1116,14 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking, Multicall {
...
 
-        // Needs to have free capacity not used for other purposes to allocate
-        require(getIndexerCapacity(_indexer) >= _tokens, "!capacity");
+        if (_tokens > 0) {
+            // Needs to have free capacity not used for other purposes to allocate
+            require(getIndexerCapacity(_indexer) >= _tokens, "!capacity");
+        } else {
+            // Allocating zero-tokens still needs to have stake
+            // Minimum indexer stake is managed by stake/unstake
+            require(stakes[_indexer].tokensSecureStake() > 0, "!stake");
```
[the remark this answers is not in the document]
abarmat: Good catch, I was looking into the same thing. Only additional change I'd do to that proposal is to use >=

tmigone (approved):

contracts/staking/Staking.sol
```diff
@@ -1175,24 +1183,26 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking, Multicall {
...
-        if (epochs > maxAllocationEpochs) {
-            require(isIndexer || isDelegator(alloc.indexer, msg.sender), "!auth-or-del");
-        } else {
+        if (epochs <= maxAllocationEpochs || alloc.tokens == 0) {
             require(isIndexer, "!auth");
         }
 
+        // Close the allocation and start counting a period to settle remaining payments from
+        // state channels.
+        allocations[_allocationID].closedAtEpoch = alloc.closedAtEpoch;
```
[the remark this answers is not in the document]
abarmat: Just did it because I added a header below it with changes related to the rebate pool, to separate allocation state update with the rest.

contracts/staking/libs/Rebates.sol
abarmat: I feel the more correct thing to do is to redefine exists() as it has more to do with having the alpha initialized, is should have always been like that. wdyt?

test/staking/allocation.test.ts
abarmat: I find it was a bunch of code to copy/paste/repeat, let's think about how we can make them clearer in a different PR, maybe we can abstract part of it in different functions as behaviors.

contracts/staking/libs/Rebates.sol
pcarranzav: Yeah that sounds good to me (and it's probably okay to handle in a different PR?)

test/staking/allocation.test.ts
pcarranzav: Yeah sounds good!

contracts/staking/libs/Rebates.sol
abarmat: I'll open an issue to tackle in a different PR / upgrade cycle.

test/lib/testHelpers.ts
tmigone: oh good catch

tmigone (approved):

pcarranzav (approved):
LGTM

PR #552, opened 2022-03-24
Arbitrum bridge + rewards distribution

pcarranzav:
# Motivation

Based on discussions that have been going on in the past few months (see this [Forum discussion](https://forum.thegraph.com/t/roadmap-to-l2-utilization/2695/1)), it seems that there is interest in the community for scaling through a Layer 2 network. Conversations between core dev teams point towards an Arbitrum Graph devnet as the most viable first step in this direction.

This devnet would require, among other things, a way to bridge GRT from L1 into L2 and back. This bridge should be the canonical way to send GRT into Arbitrum, but it should be extensible to support any protocol-specific behavior that may be necessary, both for the initial devnet deployment and the eventual "mainnet” release. This GIP covers the specification of such bridge.

GIPs for discussions in the forum:
- https://forum.thegraph.com/t/gip-0031-arbitrum-grt-bridge/3305
- https://forum.thegraph.com/t/gip-0034-the-graph-arbitrum-devnet-with-a-new-rewards-issuance-and-distribution-mechanism/3418

# Changes

- We add the following contracts L1GraphTokenGateway, BridgeEscrow, L2GraphTokenGateway and L2GraphToken
- Deployment commands in the CLI are updated to include an L2 deployment
- Configuration and address book entries for Arbitrum are added as well.
- L1Reservoir and L2Reservoir are introduced, changing the RewardsManager to pull. rewards from them instead of minting.

(Edited to add the changes from #556 and #571 as these were merged on top of this PR)


.solcover.js
```diff
@@ -1,4 +1,4 @@
-const skipFiles = ['bancor', 'ens', 'erc1056']
+const skipFiles = ['bancor', 'ens', 'erc1056', 'arbitrum', 'tests/arbitrum']
```
pcarranzav, line 1: We could still include these, and just add a bunch more tests...

contracts/l2/token/L2GraphToken.sol
```diff
@@ -0,0 +1,258 @@
...
+    // https://github.com/ethereum/EIPs/blob/master/EIPS/eip-712.md#definition-of-domainseparator
+
+    bytes32 private constant DOMAIN_TYPE_HASH =
+        keccak256(
+            "EIP712Domain(string name,string version,uint256 chainId,address verifyingContract,bytes32 salt)"
+        );
+    bytes32 private constant DOMAIN_NAME_HASH = keccak256("Graph Token");
+    bytes32 private constant DOMAIN_VERSION_HASH = keccak256("0");
+    bytes32 private constant DOMAIN_SALT =
+        0xe33842a7acd1d5a1d28f25a931703e5605152dc48d64dc4716efdae1f5659591; // Randomly generated salt
```
pcarranzav, line 44: Not sure about this, it seemed safer to use a different salt for each layer?

test/l2/l2GraphTokenGateway.test.ts
```diff
@@ -0,0 +1,364 @@
...
+          utils.defaultAbiCoder.encode(['uint256', 'bytes'], [0, []])
+        ]
+      )
+      await expect(tx).emit(l2GraphTokenGateway, 'TxToL1')
+        .withArgs(tokenSender.address, mockL1Gateway.address, 1, calldata)
+
+      // For some reason the call count doesn't work properly,
+      // and each function call is counted 12 times.
+      // Possibly related to https://github.com/defi-wonderland/smock/issues/85 ?
+      //expect(arbSysMock.sendTxToL1).to.have.been.calledOnce
```
pcarranzav, line 234: I was trying to investigate why this doesn't work, as it would be nice to explicitly check we only call this once...

pcarranzav (commented):
Some notes and open questions.

pcarranzav:
I've deployed this to Rinkeby and I'm working on some CLI commands to configure the new contracts and interact with them, but doing this in a separate PR, see #556

contracts/gateway/GraphTokenGateway.sol
```diff
@@ -0,0 +1,60 @@
...
+     */
+    function setPaused(bool newPaused) external onlyGovernorOrGuardian {
+        _setPaused(newPaused);
+    }
+
+    /**
+     * @notice Getter to access paused
+     */
+    function paused() external view returns (bool) {
+        return _paused;
```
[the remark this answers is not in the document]
pcarranzav: In [40464cc](https://github.com/graphprotocol/contracts/pull/552/commits/40464cca47f7ac0dbea2663f2488c9fdb557772c) I changed it so that controller.paused won't pause the gateways, so now this comment doesn't make much sense.

.solcover.js
pcarranzav: (I would suggest leaving this as it is for now though, as it makes little sense to test thirdparty/mock code that we don't use, and coverage would naturally test the parts that we do use)

contracts/gateway/L1GraphTokenGateway.sol
```diff
@@ -0,0 +1,258 @@
...
+     * @return Sequence number of the retryable ticket created by Inbox
+     */
+    function outboundTransfer(
+        address _l1Token,
+        address _to,
+        uint256 _amount,
+        uint256 _maxGas,
+        uint256 _gasPriceBid,
+        bytes calldata _data
+    ) external override payable notPaused nonReentrant returns (bytes memory) {
```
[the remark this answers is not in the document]
pcarranzav: I've removed the guards as I'm quite confident we don't need them (these functions are not doing any untrusted calls) but I'd appreciate someone else confirming this makes sense.

contracts/l2/token/L2GraphToken.sol
abarmat: The DOMAIN_SEPARATOR should be different anyways from the mainnet one by the call to `getChainID()` https://github.com/graphprotocol/contracts/blob/dev/contracts/disputes/DisputeManager.sol#L184

contracts/gateway/L1GraphTokenGateway.sol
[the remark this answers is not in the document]
abarmat: I agree, we are interacting with trusted ones that are not doing a recursive call.

contracts/l2/token/L2GraphToken.sol
pcarranzav: Do you think we should keep the same salt then? To me it seemed intuitively a bit cleaner to use a new random one, but I agree the chain ID should be sufficient to get different separators so it's probably irrelevant.

test/l2/l2GraphTokenGateway.test.ts
abarmat: The error looks very similar to what they reported, and the number being 12 looks really fixed. I haven't used that feature from waffle yet.

pcarranzav:
Superseded by #699 so closing

PR #555, opened 2022-03-29
Make block/epoch alignment in tests more predictable

pcarranzav:
(Temporarily targeting the pcv/fix-tests branch as this builds on #554)

Now that blocks are mined predictably in our unit tests (i.e. one block per transaction), any tests that rely on a specific number of blocks passing can break if the number of transactions changes. This was evident when I incorporated the changes from #554 into #552, see: https://github.com/graphprotocol/contracts/runs/5728132209?check_suite_focus=true - adding a transaction in the fixtures bringup caused the test to spill over into the next epoch, changing the results.

So this PR makes a few changes to make tests hopefully more predictable and robust:
- Change the `advanceToNextEpoch` function to always advance to the first block in the next epoch, rather than always advancing a full epoch
- Change several tests that advanced an epochLength to use advanceToNextEpoch instead
- In takeRewards tests that expect an amount for indexer rewards, align to the first block in an epoch and recalculate the expected rewards using the number of blocks, that is now a bit easier to calculate by just counting the number of transactions in the test code.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -125,7 +125,7 @@ const config: HardhatUserConfig = {
         mnemonic: DEFAULT_TEST_MNEMONIC,
       },
       mining: {
-        auto: true,
+        auto: false,
         interval: 30000,
       },
       hardfork: 'london',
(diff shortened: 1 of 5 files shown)


abarmat (approved):
Great improvements ⚡️

PR #564, opened 2022-05-09
Present a POI (proof of indexing) when opening an allocation [rebased]

abarmat:
This is a rebased version of https://github.com/graphprotocol/contracts/pull/544

Keeping the original branch for the audited commit hash as reference.


pcarranzav:

PR #567, opened 2022-05-12
Allow callhooks from whitelisted addresses in the Arbitrum GRT bridge

pcarranzav:
We'll need some trusted contracts to atomically sync state from L1 while sending tokens, so this allows us to do it using the bridge. The whitelist is managed by governance so it won't allow untrusted contract calls.

Example uses (TBD if we'll actually implement these exactly like this, or some other way):
- Sending newly-minted indexer rewards to L2, and at the same time updating the cached total token supply or any other parameters needed for rewards calculation.
- Migrating tokens for GNS curation signal for a subgraph, and creating the subgraph on the L2 GNS.

I'll update GIP-0031 to include this addition.

### DIFF
--- a/contracts/gateway/L1GraphTokenGateway.sol
+++ b/contracts/gateway/L1GraphTokenGateway.sol
@@ -30,6 +30,8 @@ contract L1GraphTokenGateway is GraphTokenGateway, L1ArbitrumMessenger {
     address public l2Counterpart;
     // Address of the BridgeEscrow contract that holds the GRT in the bridge
     address public escrow;
+    // Address of the L1 Reservoir that is the only sender allowed to send extra data
+    mapping(address => bool) public callhookWhitelist;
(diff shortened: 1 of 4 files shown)


pcarranzav:
Looks like the Arbitrum team is re-adding the depositAndCall functionality to their reference implementation: https://github.com/OffchainLabs/arbitrum/pull/2235 so it might be worth revisiting this when that's ready, to make it compatible if that interface works for our use case.

pcarranzav:
After chatting about it with @abarmat, I'll merge this into the main bridge PR as it is, so that it's included in our upcoming audit. If/when the Offchain Labs team releases the depositAndCall feature, we can revisit and upgrade the bridge if it's beneficial.
I still have to post the update to GIP-0031 to include this feature, so I'll do that soon.

PR #568, opened 2022-05-16
feat: make GNS signal transferable

tmigone:
See https://forum.thegraph.com/t/gip-0032-make-gns-signal-transferrable/3338

### Motivation

From GIP32:

> Curators can mint curation shares from a Subgraph Deployment directly on the Curation contract and get fully transferrable ERC20 shares. However, when they curate a Subgraph at the GNS level, the GNS contract is the one keeping all the signal shares under its control. The subgraph curators cannot transfer those shares because the GNS is not exposing any function to do that.

### Changes

This PR adds a `transferSignal` function to the GNS contract that allows a curator to transfer ownership of their `nSignal` to another address.

### DIFF
--- a/contracts/discovery/GNS.sol
+++ b/contracts/discovery/GNS.sol
@@ -212,10 +212,9 @@ contract GNS is GNSV2Storage, GraphUpgradeable, IGNS, Multicall {
      * @param _subgraphNFT Address of the ERC721 contract
      */
     function _setSubgraphNFT(address _subgraphNFT) private {
-        require(
-            _subgraphNFT != address(0) && Address.isContract(_subgraphNFT),
-            "NFT must be valid"
-        );
+        require(_subgraphNFT != address(0), "NFT address cant be zero");
+        require(Address.isContract(_subgraphNFT), "NFT must be valid");
+
         subgraphNFT = ISubgraphNFT(_subgraphNFT);
         emit SubgraphNFTUpdated(_subgraphNFT);
     }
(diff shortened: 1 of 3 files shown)


contracts/discovery/GNS.sol
```diff
@@ -86,6 +86,16 @@ contract GNS is GNSV2Storage, GraphUpgradeable, IGNS, Multicall {
         uint256 tokensReceived
     );
 
+    /**
+     * @dev Emitted when a curator transfers signal.
+     */
+    event SignalTransfered(
```
pcarranzav, line 92: nit/typo: should be `SignalTransferred`

contracts/discovery/GNS.sol
```diff
@@ -86,6 +86,16 @@ contract GNS is GNSV2Storage, GraphUpgradeable, IGNS, Multicall {
...
+        uint256 indexed subgraphID,
+        address indexed from,
+        address indexed to,
+        uint256 nSignalTransfered
```
pcarranzav, line 96: nit/typo: should be `nSignalTransferred`

abarmat (dismissed):
This is ready for an audit

tmigone:
Latest commit https://github.com/graphprotocol/contracts/pull/568/commits/09f40d160b2a65e9672ca005fa85e329bd52d3e4 addresses audit issues N-22 and L-11

abarmat (approved):

PR #569, opened 2022-05-16
Rewards: fix calculation by using a snapshot of token supply at the time that signal was last updated

pcarranzav:
Current rewards calculation tries to follow a curve `p*r^(t-b) - p`, where `p` is token supply, `r` is issuance rate plus one, `t` is current block and `b` is the block when signal was last updated on any subgraph.

For this to follow the expected curve, however, `p` must correspond to the token supply at block `b`, and not to the time-varying instant supply. This PR fixes the calculation by adding a token supply snapshot, and adds two regression tests that are fixed by this PR and that exemplify the consequences of this bug.

Note that this also changes the `RewardsManager.initialize()` function by removing the call to set issuance rate, since we now need the address of the `GraphToken` contract (to get the supply snapshot) that is not available at initialization time.

### DIFF
--- a/config/graph.mainnet.yml
+++ b/config/graph.mainnet.yml
@@ -103,7 +103,9 @@ contracts:
     proxy: true
     init:
       controller: "${{Controller.address}}"
-      issuanceRate: "1000000012184945188" # per block increase of total supply, blocks in a year = 365*60*60*24/13
+    calls:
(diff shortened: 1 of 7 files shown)


abarmat (dismissed):
Looking good! 🥇

abarmat (approved):

PR #571, opened 2022-05-17
Arbitrum devnet with rewards distribution

pcarranzav:
This PR adds a full deployment to the Arbitrum One layer 2, and refactors rewards distribution so that they can be periodically dripped on L1. A fraction of the minted rewards can be sent to L2 (which would initially be 0 in the devnet phase).

Full GIP pending.

This PR temporarily targets the `pcv/arb-bridge-cli` branch as it depends on #552 and #556 being merged first. It also includes the changes from #569 as I'm assuming that will be merged and deployed before this.


pcarranzav:
Marking this as ready for review, but it depends on #552 and #556 being merged first. I'll also be working on writing the GIP for this soon.
This has **not** been tested on a testnet yet, only with the unit tests provided in the PR.

abarmat (approved):

PR #572, opened 2022-05-17
Tmigone/update config task

tmigone:


package.json
```diff
@@ -81,9 +81,9 @@
     "compile": "hardhat compile",
     "deploy": "scripts/predeploy && hardhat migrate",
     "deploy-ganache": "yarn deploy -- --force",
-    "deploy-ganache-manual": "yarn deploy -- --network ganache --force",
-    "deploy-hardhat": "yarn deploy -- --network hardhat --force",
-    "deploy-rinkeby": "yarn deploy -- --force --network rinkeby",
+    "deploy-ganache-manual": "yarn deploy -- --force --network ganache",
+    "deploy-hardhat": "yarn deploy -- --force --network hardhat",
```
abarmat, line 85: What if --graph-config is not passed? is it storing it in any place by default or just not storing it?

tasks/deployment/config.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// GraphToken
+// TODO > 'initialSupply' not on chain
+// const graphToken: Contract = {
+//   name: 'GraphToken',
+//   initParams: [],
+// }
+
+const curation: Contract = {
```
abarmat, line 38: You could import the ABIs JSON from the `build/abi` folder to avoid copying them over.

abarmat (commented):
Very clever to fetch config from on-chain data, it's a nice addition to the existing commands. I've just left a few notes.

tasks/deployment/config.ts
tmigone: @abarmat I thought about using ABIs but how would you filter out all the stuff that's not related to init params? You'd still need to hardcode a list right?

abarmat: @tmigone I get it, you are using the list to know what to scan, all good!

abarmat (approved):

abarmat:
@tmigone is this ready to merge?

tmigone:
@abarmat the second commit is pending review. Not sure why Pushing  a new commit didn’t  invalidate the original approval

abarmat:
Oh, I see

abarmat (approved):
🥇

PR #575, opened 2022-05-23
Add GraphCurationToken to addressbook

tmigone:
`GraphCurationToken` addresses were missing from the address book on both mainnet and rinkeby. Code hashes created with [this](https://github.com/tmigone/hello-eth/blob/297998981dfa189ec9dccd4649a75f21eae295f7/scripts/codeHash.ts) script.


abarmat (approved):
Good catch!

Issue #576, opened 2022-05-23
GNS tests fail if subgraphNFT's tokenURI has leading zeros


tmigone:
Will work on reproducing the issue, in the meantime here are logs from a recent failed PR check:

```

  1) GNS
       NFT descriptor
         without token descriptor:

      AssertionError: expected '0x00fae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a' to equal '0xfae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a'
      + expected - actual

      -0x00fae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a
      +0xfae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a
      
      at Context.<anonymous> (test/gns.test.ts:1042:27)

  2) GNS
       NFT descriptor
         without token descriptor and baseURI:

      AssertionError: expected 'ipfs://0x00fae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a' to equal 'ipfs://0xfae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a'
      + expected - actual

      -ipfs://0x00fae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a
      +ipfs://0xfae00cac62e26a062ca75eb35392df730f5e6c84aa36aa7d199b4bd713f73a
      
      at Context.<anonymous> (test/gns.test.ts:1055:39)
```

tmigone:
Another instance of this happening: https://github.com/graphprotocol/contracts/actions/runs/3567393752/jobs/5995037836

tmigone:
Happened again: https://github.com/graphprotocol/contracts/actions/runs/4006618984/jobs/6878385117

PR #577, opened 2022-05-24
fix: ensure subgraphNFT handles metadata with leading zeros correctly

tmigone:
### Motivation

The tokenURI being reported by SubgraphNFT contract using the `tokenURI` function is incorrect when the subgraph metadata hash contains leading zeroes. #576 describes the problem in detail.

### Changes

- Added test to check for this special case where subgraph metadata hash is of the form `0x00....`
- Fix in SubgraphNFT contract

Closes: #576


pcarranzav:
Since we won't be able to deploy this fix, should we close this and leave the issue open as a known bug? (We can reopen if we ever get a chance to deploy a new SubgraphNFT for whatever reason)

pcarranzav:
^ closing based on the rationale above

PR #581, opened 2022-06-05
feat: use Gelato rewards for drip (WIP, experimental, doesn't work yet)

pcarranzav:
Playing a bit with how we could potentially integrate with Gelato to provide a reward for the reservoir drip (introduced in #571). The L1 side looks pretty straightforward, but the L2 side wouldn't work at all, so this needs some thought 🤔


pcarranzav:
Closing for now, back to the drawing board and we have other ideas

PR #582, opened 2022-06-05
feat: keeper reward for reservoir drip through token issuance

pcarranzav:
WIP as I still need to fix the tests.

This complements #571 adding a reward for whoever called drip(), to incentivise calling it and offset the gas costs. The reward is produced through additional token issuance that grows linearly with the number of blocks, after a minimum number of blocks since the last drip have passed.

Note the reward is credited on L1 if the l2RewardsFraction is set to 0. If l2RewardsFraction is nonzero, the reward will be credited in L2, and part of it will be given to the address that initiated the retryable ticket transaction, so that if a lazy or malicious keeper does not redeem the tx in L2, there is incentive for someone else to jump in and redeem it.


pcarranzav:
(Fixed the tests, but I still need to add a test for distributing the keeper reward in L2)

pcarranzav:
Rebased on top of #624 to incorporate all the audit fixes. This caused lot of merge conflicts that needed to be solved...

abarmat (approved):
👍

PR #583, opened 2022-06-07
chore: use automine for local deployments

abarmat:
Improve deployment speed for local tests with automine while keeping 13s for real condition testing.


tmigone (approved):
lgtm!

PR #585, opened 2022-06-27
feat: transfer of subgraphs and curation to L2 through GNS

pcarranzav:
GIP pending.

This PR would be part of the third stage of The Graph's migration to Arbitrum.

It does the following:
- Add an L1GNS and L2GNS (replacing GNS on each layer), with helper functions to send a subgraph from L1 to L2 with the owner's curation signal.
- L1GNS can also send specific curators' balances for migrated subgraphs to L2 through the bridge.
- L2Curation replaces Curation in L2, removing the bonding curve logic as reserve ratio in L2 is meant to be always 1 (i.e. a flat curve).
- QA improvements across affected files: using explicit imports, re-ordering functions to follow the Solidity style guide, and some gas efficiency changes (e.g. using `!= 0` instead of `> 0` where appropriate)

See GIP-0046 for details: https://forum.thegraph.com/t/gip-0046-l2-migration-helpers/4023


contracts/l2/discovery/L2GNS.sol
```diff
@@ -0,0 +1,315 @@
...
+        // The subgraph will be disabled until finishSubgraphMigrationFromL1 is called
+        subgraphData.disabled = true;
+        subgraphData.nSignal = _nSignal;
+
+        migratedData.tokens = _tokens;
+        migratedData.subgraphReceivedOnL2BlockNumber = block.number;
+
+        // Mint the NFT. Use the subgraphID as tokenID.
+        // This function will check the if tokenID already exists.
+        _mintNFT(_subgraphOwner, _subgraphID);
```
tmigone, line 294: Shouldn't the NFT mint happen on the `finishSubgraphMigrationFromL1` function? 

I can see how it could be done in both places but doing it here increases the callhook complexity (plus the NFT could be worthless if the migration is not finished).

contracts/l2/discovery/L2GNS.sol
```diff
@@ -0,0 +1,315 @@
...
+ */
+contract L2GNS is GNS, L2GNSV1Storage, IL2GNS {
+    using SafeMathUpgradeable for uint256;
+
+    /// The amount of time (in blocks) that a subgraph owner has to finish the migration
+    /// from L1 before the subgraph can be deprecated: 1 week
+    uint256 public constant FINISH_MIGRATION_TIMEOUT = 50400;
+
+    /// @dev Emitted when a subgraph is received from L1 through the bridge
+    event SubgraphReceivedFromL1(uint256 indexed _subgraphID);
```
tmigone, line 36: What do you think about adding the subgraph owner and tokens?

contracts/l2/discovery/L2GNS.sol
```diff
@@ -0,0 +1,315 @@
...
+            _balance
+        );
+        migratedData.curatorBalanceClaimed[_curator] = true;
+        emit CuratorBalanceClaimed(_subgraphID, _curator, _beneficiary, _balance);
+    }
+
+    /**
+     * @notice Publish a new version of an existing subgraph.
+     * @dev This is the same as the one in the base GNS, but skips the check for
+     * a subgraph to not be pre-curated, as the reserve ration in L2 is set to 1,
```
tmigone, line 198: typo 

```suggestion
     * a subgraph to not be pre-curated, as the reserve ratio in L2 is set to 1,
```

contracts/l2/discovery/L2GNS.sol
pcarranzav: We do it here so that we can then use `onlySubgraphAuth` in `finishSubgraphMigrationFromL1`. You're right that it's not great to increase complexity in the callhook, but I thought this was still better than persisting the owner some other way and doing a custom ownership check

tmigone: Ah I missed that modifier. In that case I think adding a comment noting that would be helpful.

contracts/l2/discovery/L2GNS.sol
pcarranzav: Good point, I'll fix it in #764

contracts/l2/discovery/L2GNS.sol
```diff
@@ -0,0 +1,315 @@
...
+     */
+    function finishSubgraphMigrationFromL1(
+        uint256 _subgraphID,
+        bytes32 _subgraphDeploymentID,
+        bytes32 _subgraphMetadata,
+        bytes32 _versionMetadata
+    ) external override notPartialPaused onlySubgraphAuth(_subgraphID) {
+        IL2GNS.SubgraphL2MigrationData storage migratedData = subgraphL2MigrationData[_subgraphID];
+        SubgraphData storage subgraphData = _getSubgraphData(_subgraphID);
+        // A subgraph
```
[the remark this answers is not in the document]
pcarranzav: Probably, I'll fix it in #764

contracts/l2/discovery/L2GNS.sol
pcarranzav: Fixing in #764

contracts/l2/discovery/L2GNS.sol
pcarranzav: Good point, I'll fix it in #764

tmigone (dismissed):

abarmat (dismissed):

abarmat (approved):

PR #614, opened 2022-07-13
chore: add a confirmation step before deploying from CLI

abarmat:


tmigone (approved):

tmigone:
Added a commit with a few more commands

PR #620, opened 2022-07-14
fix: remove unneeded ERC20Upgradeable inheritance [N-17]

pcarranzav:


pcarranzav:
Broken test seems to be an instance of https://github.com/graphprotocol/contracts/issues/576 - so retrying

PR #624, opened 2022-07-14
fix: check before adding/removing whitelisted addresses [N-23]

pcarranzav:


abarmat (commented):
Just nitpicking a bit, this function might also have been something like:
```
function setCallhookWhitelist(address _target, bool _enabled) external onlyGovernor
```
But your version is more explicit, it's good 👍 

In the spirit of failing loudly, I'd revert if the callhook is already whitelisted like:
```
require(callhookWhitelist[_newWhitelisted] == false, "WHITELIST_SET");
```
I try to avoid a transaction that is successful and doesn't do a state change.

What do you think?

pcarranzav:
> Just nitpicking a bit, this function might also have been something like:
> 
> ```
> function setCallhookWhitelist(address _target, bool _enabled) external onlyGovernor
> ```
> 
> But your version is more explicit, it's good 👍

Cool, yeah, I wasn't sure which way was better, I wonder if any of the options is a bit more gas-efficient, but in the end this won't be called very often so probably no big deal either way...

> In the spirit of failing loudly, I'd revert if the callhook is already whitelisted like:
> 
> ```
> require(callhookWhitelist[_newWhitelisted] == false, "WHITELIST_SET");
> ```
> 
> I try to avoid a transaction that is successful and doesn't do a state change.
> 
> What do you think?

Good point, I wasn't sure what to do for this, since it's technically okay to try to set the whitelisted address twice, and I liked the function being idempotent, but in this case it might be safer to revert as it might be a sign that the user did something wrong...

pcarranzav:
Amended to incorporate @abarmat's suggestion to revert rather than silently skip

abarmat (dismissed):

abarmat (approved):

PR #626, opened 2022-07-15
chore: add hardhat task to verify all contracts with etherscan

tmigone:


abarmat:
Very helpful for new full protocol deployments.

PR #642, opened 2022-07-22
feat: improve account management

abarmat:
### Changes

- Allow to use a mnemonic or single private key 
- Choose whether to use hosted or remote keys when forking


tmigone (approved):
🚀

PR #653, opened 2022-07-28
fix: rewards manager e2e config test

tmigone:
PR #569 broke one of the rewards manager e2e config tests as it removed an initialization value from the initializer and instead is setting it with a function call. 

Connected to: #654


abarmat (approved):
Great fix!

PR #661, opened 2022-07-28
chore: bump npm package version

tmigone:


abarmat (commented):
I think 1.14.0 is free for use

tmigone:
That was my bad, I misunderstood the way we worked around the side branches like the special testnet for product team. 

I bumped to `1.14.0-testnet-product` for the product testnet which I shouldn't have, should have just published as `1.13.0-testnet-product`. I'll change this one to `1.14.0` so we are back on track.

abarmat (approved):

PR #674, opened 2022-08-05
feat: add support for local nitro networks

tmigone:
### Motivation

This PR adds the local nitro networks L1 and L2 to hardhat and fixes a few bugs to allow deploying the protocol to those networks.

### Changes
- Add local nitro networks (L1 and L2) to hardhat
- Fail gracefully if the network doesn't support `evm_setAutomine` on local networks. Nitro uses geth in dev mode which already has automine behaviour and doesn't support the RPC call.

### DIFF
--- a/cli/commands/migrate.ts
+++ b/cli/commands/migrate.ts
@@ -72,7 +72,7 @@ export const migrate = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<vo
 
   if (chainId == 1337) {
     allContracts = ['EthereumDIDRegistry', ...allContracts]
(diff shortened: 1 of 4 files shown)


cli/commands/migrate.ts
```diff
@@ -72,7 +72,7 @@ export const migrate = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<vo
 
   if (chainId == 1337) {
```
abarmat, line 73: Once we move the list of contracts to deploy to a config file we won't need this if/case

abarmat (approved):

PR #678, opened 2022-08-09
feat: add calldata parameter to send to l2 command

abarmat:


cli/commands/bridge/to-l2.ts
```diff
@@ -61,6 +62,7 @@ export const sendToL2 = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<v
   const amount = toGRT(cliArgs.amount)
   const recipient = cliArgs.recipient ? cliArgs.recipient : cli.wallet.address
   const l2Dest = await gateway.l2Counterpart()
+  const calldata = cliArgs.calldata ?? '0x'
```
pcarranzav, line 65: TIL about the `??` operator, nice

pcarranzav (approved):
LGTM

cli/commands/bridge/to-l2.ts
```diff
@@ -161,7 +163,7 @@ export const redeemSendToL2 = async (cli: CLIEnvironment, cliArgs: CLIArgs): Pro
 }
 
 export const sendToL2Command = {
-  command: 'send-to-l2 <amount> [recipient]',
+  command: 'send-to-l2 <amount> [recipient] [calldata]',
   describe: 'Perform an L1-to-L2 Graph Token transaction',
   builder: (yargs: Argv): Argv => {
     return yargs
```
pcarranzav, line 169: Oh, actually, I think you might need to add the calldata option below here as well? i.e.

```
.positional('calldata', { demandOption: false })
```

abarmat: It's only needed if you want to describe it or pass any special argument. I'll do it and add descriptions.

PR #679, opened 2022-08-11
some improvements in config files

pcarranzav:


config/graph.arbitrum-localhost.yml
```diff
@@ -48,6 +48,8 @@ contracts:
     proxy: true
     init:
       controller: "${{Controller.address}}"
+    calls:
+      - fn: "syncAllContracts"
   EpochManager:
```
tmigone, line 53: Missed one here 🔍

config/graph.arbitrum-one.yml
```diff
@@ -48,6 +48,8 @@ contracts:
     proxy: true
     init:
       controller: "${{Controller.address}}"
+    calls:
+      - fn: "syncAllContracts"
   EpochManager:
```
tmigone, line 53: 🔍

tmigone (changes requested):

config/graph.arbitrum-one.yml
pcarranzav: good catch, but I actually left the EpochManagers out on purpose, since they don't do any external calls... (I basically copied the list from test/lib/fixtures.ts)

abarmat (approved):

test/lib/fixtures.ts
```diff
@@ -172,7 +172,6 @@ export class NetworkFixture {
     await staking.connect(deployer).setSlasher(slasherAddress, true)
     await gns.connect(deployer).approveAll()
     if (isL2) {
-      await grt.connect(deployer).addMinter(l2GraphTokenGateway.address)
```
abarmat, line 175: Can you remind me what we removed the L2 bridge to be a minter in the fixture?

pcarranzav: Because we are already using the `L2GraphToken.setGateway` function to add it as the gateway minter, that can mint and burn through the bridgeMint and bridgeBurn functions.

abarmat (approved):

config/graph.arbitrum-one.yml
```diff
@@ -53,10 +55,8 @@ contracts:
   L2GraphToken:
     proxy: true
     init:
-      governor: "${{Env.deployer}}"
+      owner: "${{Env.deployer}}"
```
tmigone, line 58: Same :D

tmigone (changes requested):
Small nit with the`L2GraphToken` config. It can be simplified from:

```
  L2GraphToken:
    proxy: true
    init:
      owner: "${{Env.deployer}}"
    calls:
      - fn: "renounceMinter"
      - fn: "transferOwnership"
        owner: *governor
```

to this:

```
  L2GraphToken:
    proxy: true
    init:
      owner: *governor
```

config/graph.arbitrum-goerli.yml
```diff
@@ -53,10 +55,8 @@ contracts:
   L2GraphToken:
     proxy: true
     init:
-      governor: "${{Env.deployer}}"
+      owner: "${{Env.deployer}}"
```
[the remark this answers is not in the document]
pcarranzav: Hmm but this will actually leave the governor as a minter, whereas the other approach leaves no minters enabled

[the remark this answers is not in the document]
tmigone: oh good point

tmigone (approved):

PR #685, opened 2022-08-17
fix: don't include keeper reward twice when computing what to send to L2 [H-03] [L-03]

pcarranzav:
Also prevent the drip from succeeding when the amount to mint is zero.


abarmat (approved):

abarmat:
👀

PR #687, opened 2022-08-19
refactor: send tokens to L2

abarmat:
### Changes

- Move estimate gas for L2 ticket to a function for reuse
- Move some functions to it's own cross-chain library
- Convert gas params inputs into BigNumber
- Ensure allowance and skip if already done


cli/commands/bridge/to-l2.ts
```diff
@@ -49,95 +44,75 @@ export const sendToL2 = async (cli: CLIEnvironment, cliArgs: CLIArgs): Promise<v
...
-    logger.error('Auto redeem failed')
-    logger.error(e)
-    logger.error('You can re-attempt using redeem-send-to-l2 with the following txHash:')
-    logger.error(receipt.transactionHash)
+  const txData = utils.defaultAbiCoder.encode(['uint256', 'bytes'], [maxSubmissionCost, calldata])
+  const txParams = [l1GRT.address, recipient, amount, maxGas, gasPriceBid, txData]
+  const txReceipt = await sendTransaction(cli.wallet, l1Gateway, 'outboundTransfer', txParams)
+
+  // get l2 ticket status
+  if (txReceipt.status == 1) {
```
pcarranzav, line 103: Should we fail more noisily if `txReceipt.status` is not 1?

cli/cross-chain.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const estimateRetryableTxGas = async (
+  l1Provider: providers.Provider,
+  l2Provider: providers.Provider,
+  gatewayAddress: string,
+  l2Dest: string,
+  depositCalldata: string,
+  opts: L2GasParams,
+): Promise<L2GasParams> => {
+  const autoEstimate = opts && (!opts.maxGas || !opts.gasPriceBid || !opts.maxSubmissionCost)
```
pcarranzav, line 37: Is the `opts &&` needed here? I think typescript won't allow it to be null anyways (unless you mark it as optional in the function definition), but in any case if opts is nullish we would want to auto-estimate, wouldn't we?

tmigone:
Changes cherry picked onto #704

PR #694, opened 2022-08-23
test: optimize advanceBlocks by using regular mining up to 10 blocks

pcarranzav:
This is (hopefully) a slight improvement over what was introduced in #688 - where we switched to using hardhat_mine whenever we want to advance a number of blocks. It seems this is a bit slower than just mining blocks iteratively when the number of blocks is low, so here we use regular mining for up to 10 blocks, and hardhat_mine if it's more than that.

(the optimal number might be more or less than 10, and probably depends on the machine... but we can tune it to find something that's reasonable)

### DIFF
--- a/test/lib/testHelpers.ts
+++ b/test/lib/testHelpers.ts
@@ -76,7 +76,15 @@ export const advanceBlockTo = async (blockNumber: string | number | BigNumber): 
 }
 
 export const advanceBlocks = async (blocks: string | number | BigNumber): Promise<void> => {
-  await provider().send('hardhat_mine', [hexValue(BigNumber.from(blocks))])
+  const blocksBN = BigNumber.from(blocks)
+  const maxIterativeBlocks = BigNumber.from(10)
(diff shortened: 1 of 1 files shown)


pcarranzav:
Just did a quick benchmark, advancing blocks with evm_mine takes about 1.4 ms per block, whereas advancing with hardhat_mine takes about 2.5 ms per call, independent of the number of blocks. This makes this optimization unnecessary (we could set the magic number to 2, but it won't make much difference in our total test times). So closing this, we can reopen if someone finds something different.

abarmat:
I wonder why this is not improving the performance, maybe we should report our finding to the Hardhat team

pcarranzav:
But hardhat_mine _is_ improving our performance, particularly when advancing several blocks at a time (e.g. the change in #688 would simply not work without it). I think our total test time isn't affected much simply because advancing blocks is not the main bottleneck (except for the one in #688).

abarmat:
I miss read, the optimization was in the other PR and works 👍

PR #696, opened 2022-08-25
feat(gre): add hardhat-secure-accounts to GRE

tmigone:
### Motivation

Add hardhat-secure-accounts support to GRE.

### Changes

Account management methods (`getDeployer`, `getTestAccounts` and `getNamedAccounts`) are now hooked up to hardhat-secure-accounts plugin by default. This can be disabled using the flag `disableSecureAccounts`, see GRE readme for more details. 

Example:

```js
// Without secure accounts
> const graph = hre.graph({ disableSecureAccounts: true })
> const deployer = await g.l1.getDeployer()
> deployer.address
'0xBc7f4d3a85B820fDB1058FD93073Eb6bc9AAF59b'


// With secure accounts
> const graph = hre.graph()
> const deployer = await g.l1.getDeployer()
== Using secure accounts, please unlock an account for L1(goerli)
Available accounts:  goerli-deployer, arbitrum-goerli-deployer, rinkeby-deployer, test-mnemonic
Choose an account to unlock (use tab to autocomplete): test-mnemonic
Enter the password for this account: ************
> deployer.address
'0x90F8bf6A479f320ead074411a4B0e7944Ea8c9C1'
```

**Note**
Shuffled the code a bit which makes this seem like a bigger change than it is. Renamed `helpers/network.ts` --> `helpers/chain.ts` and also added an unrelated `helpers/network.ts` which makes GitHub diff viewer to go wild.


abarmat (changes requested):
I'd maybe add a reference to do the `hh accounts new` if they want to setup a new account so they don't need to look at the reference from the plugin.

The plugin is throwing an error when then accounts key in the hardhat.config.js entry for a network returns a string, which is a valid value as in `remote`

tmigone:
Fixed the error!

About the documentation reference, the GRE README.md mentions the plugin and links to it already, do you think we should add more?

abarmat (approved):
Works!

PR #697, opened 2022-08-30
[Experimental] Arbitrum bridge + rewards distribution: alternative approach

pcarranzav:
Trying a different approach to the one from GIP-0034: Minting on L2 and not using a reservoir drip. For now just making the changes to compare and make sure we're choosing the best solution. WIP.


pcarranzav:
Replaced by #700

PR #676, opened 2022-08-05
feat(e2e): add e2e deployment tests for L2s

tmigone:
This PR adds task changes and e2e deployment tests for L1/L2. This allows us to validate steps 1.1 through 1.12 with tests (except 1.9 & 1.10 which are pending GRE support for multiple layers).

### Changes
- Add `chainId` as optional parameter when initializing GRE. This gives some flexibility when choosing where to connect to.
- hh task `e2e` no longer runs scenarios, they can be run separately if needed with `e2e:scenario`.
- hh task `migrate:unpause` will unpause the bridge as well as the protocol.
- Fix bug where `migrate:ownership` would fail if run on L2.
- Add L2 e2e deployment tests
- Add bridge (L1 & L2) e2e tests

PR #704, opened 2022-09-05
Sync: l2-testnet > pcv/l2-bridge

tmigone:
This PR cherry picks commits from https://github.com/graphprotocol/contracts/pull/659 and https://github.com/graphprotocol/contracts/pull/676. Before reviewing and merging, I suggest we merge #703, then rebase `pcv/l2-bridge` with those changes before merging this one:
- [x] `pcv/l2-bridge` rebased
- [x] #703 merged

### Changes
- Some CLI improvements for L1/L2 interaction
- Add hardhat task to sync controller contracts
- Add e2e tests for L2
- Add support for e2e tests to run on nitro testnodes


cli/commands/migrate.ts
```diff
@@ -169,7 +169,15 @@ export const migrate = async (
   logger.info(`Sent ${nTx} transaction${nTx === 1 ? '' : 's'} & spent ${EtherSymbol} ${spent}`)
 
   if (chainId == 1337) {
-    await (cli.wallet.provider as providers.JsonRpcProvider).send('evm_setAutomine', [autoMine])
+    await setAutoMine(cli.wallet.provider as providers.JsonRpcProvider, autoMine)
```
pcarranzav, line 172: Oooh now I remember what we were trying to do with the automine settings here. The idea was to set it to true above (before starting the migration), and then this step at the end would set it back to false when we're done, because at the time we were disabling automine by default when running a local hardhat node.

Now that we have the autoMine task flag, what this is actually doing is setting up the autoMine for _after_ the migration - the migration itself is always run with automine = true.

So I wonder if the best thing to do here would be to simply restore whatever automine setting was set before we start the migration?

tmigone: Yes I think I agree that's the most sensible thing to do, I'll do that! Do you know why the hardhat network sets autoMine to false in the network config?

pcarranzav:
Since we're including the CLI improvements, it might make sense to cherry pick #687 here as well?

tmigone:
Good idea, I'll close the other PR.

pcarranzav (approved):
LGTM

PR #713, opened 2022-09-21
feat(cli): add publish and mint using multicall

abarmat:


cli/helpers.ts
```diff
@@ -46,20 +49,24 @@ export class IPFS {
 export const pinMetadataToIPFS = async (
   ipfs: string,
   type: string,
-  path?: string, // Only pass path or metadata, not both
+  filepath?: string, // Only pass path or metadata, not both
```
tmigone, line 52: I was curious if there was a ts trick to require only one of two properties:

```
type RequireOnlyOne<T, Keys extends keyof T = keyof T> =
    Pick<T, Exclude<keyof T, Keys>>
    & {
        [K in Keys]-?:
            Required<Pick<T, K>>
            & Partial<Record<Exclude<Keys, K>, undefined>>
    }[Keys]
    ```

I think we can stick with your code 😆

tmigone (approved):
Looks good. My only gripe is that I'd love to stop adding new commands to the cli and use hardhat tasks instead, but we can migrate them later as well.

PR #717, opened 2022-09-22
feat(e2e): deploy and configure both layers with the e2e script

tmigone:
### Motivation

Previously the e2e testing framework would allow running tests on either an L1 or L2 deployment, with cross-chain messages not being a possibility.

This PR updates the e2e framework so we can now deploy both L1 and L2, configure them fully (including bridge) and have the ability to send L1 <> L2 messages. This requires setting up Arbitrum Nitro local testnodes using the Edge & Node patched fork: https://github.com/edgeandnode/nitro. A new simple scenario that uses the bridge to send GRT from L1 to L2 is also included.

Also updating the GitHub CI action to make use of this, a tl;dr of what the action will now be doing:
- Start an L1 node (geth) and an L2 sequencer connected to it.
- Deploy arbitrum's contracts to L1/L2
- Deploy The Graph protocol to L1/L2
- Configure the protocol (L1, L2, bridge)
- Run e2e tests on both L1 and L2 to check everything looks good
- Run e2e scenarios on both L1 and L2
   - Publish subgraphs, add signal
   - Open allocations
   - Send GRT from L1 to L2
   - Close allocations (Not working yet, need to figure out a way of advancing time)

Coming up next:
- Fully test the bridge configuration (including cross checking the addresses on both ends, contract ownership, auth, etc).
- Test more operations
   - Send GRT to L2 without ticket autoredeem
   - Send GRT with additional calldata
   - Send GRT from L2 to L1  
 
### Changes

- Update `@arbitrum/sdk` to `v3.0.0-rc1`
- Improve some deployment tasks
- Add some new nitro specific tasks
- Update e2e testing framework to allow running L1 and L2 tests on a configured L1/L2 deployment (including bridge)
- Update github action to run the L1/L2 e2e tests on nitro local testnodes.
- Added a new e2e scenario to send GRT from L1 to L2

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -25,9 +25,7 @@ jobs:
         run: |
           git clone https://github.com/edgeandnode/nitro
           pushd nitro
-          git checkout ci
           git submodule update --init --recursive
           ./test-node.bash --init --no-blockscout --detach
           popd
-          NETWORK=localnitrol1 ADDRESS_BOOK=addresses.json GRAPH_CONFIG=config/graph.localhost.yml yarn test:e2e
-          NETWORK=localnitrol2 ADDRESS_BOOK=addresses.json GRAPH_CONFIG=config/graph.arbitrum-localhost.yml yarn test:e2e
+          L1_NETWORK=localnitrol1 L2_NETWORK=localnitrol2 yarn test:e2e
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,5 @@ coverage.json
 
 # Local test files
 addresses-local.json
-arbitrum-addresses-local.json
 localNetwork.json
+arbitrum-addresses-local.json
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ coverage.json
 # Local test files
 addresses-local.json
 arbitrum-addresses-local.json
+localNetwork.json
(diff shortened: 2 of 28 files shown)


abarmat (commented):
I tried running the e2e locally, it worked OK when doing L1 only, then I set the L2_NETWORK variable to localhost and got some errors, I'm not sure how to set it to make it run on a local env.

I got an error when doing `L1_NETWORK=localnitrol1 L2_NETWORK=localnitrol2 yarn test:e2e` -> `ProviderError: sender doesn't have enough funds to send tx. The max upfront cost is: 31501500000000 and the sender's account only has: 0` - should I run the nitro setup separately?

tmigone:
> I tried running the e2e locally, it worked OK when doing L1 only, then I set the L2_NETWORK variable to localhost and got some errors, I'm not sure how to set it to make it run on a local env.

The script makes the assumption that you are either running only on L1 or running both on L1/L2. I think running ONLY on L2 using a hardhat node (which is what you get with `localhost`) is not really representative, but I'll push a fix now to allow it.

> I got an error when doing L1_NETWORK=localnitrol1 L2_NETWORK=localnitrol2 yarn test:e2e -> ProviderError: sender doesn't have enough funds to send tx. The max upfront cost is: 31501500000000 and the sender's account only has: 0 - should I run the nitro setup separately?

Yes you need to setup the nitro testnodes separately. You can check the `TESTING.md` file for instructions.

abarmat (approved):
I managed to get it running by looking at the instructions

pcarranzav (changes requested):
Looking good, but let's block the merge until the base branch is done with audits

PR #719, opened 2022-09-28
L2 bridge: safer callhook interface

pcarranzav:
- Make the L2 gateway functions non-reentrant for additional safety
- Change the callhooks to only call an `onTokenTransfer` function, to limit the potential impact of a compromised whitelisted L1 sender

### DIFF
--- a/contracts/gateway/ICallhookReceiver.sol
+++ b/contracts/gateway/ICallhookReceiver.sol
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+/**
+ * @title Interface for contracts that can receive callhooks through the Arbitrum GRT bridge
+ * @dev Any contract that can receive a callhook on L2, send through the bridge from L1, must
+ * be whitelisted by the governor, but also implement this interface that contains
+ * the function that will actually be called by the L2GraphTokenGateway.
+ */
+pragma solidity ^0.7.6;
+
+interface ICallhookReceiver {
+    /**
+     * @dev Receive tokens with a callhook from the bridge
+     * @param _from Token sender in L1
+     * @param _amount Amount of tokens that were transferred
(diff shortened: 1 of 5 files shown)


contracts/l2/gateway/L2GraphTokenGateway.sol
```diff
@@ -238,15 +241,9 @@ contract L2GraphTokenGateway is GraphTokenGateway, L2ArbitrumMessenger {
                 bytes memory gatewayData;
                 (gatewayData, callhookData) = abi.decode(_data, (bytes, bytes));
             }
-            bool success;
-            // solhint-disable-next-line avoid-low-level-calls
-            (success, ) = _to.call(callhookData);
             // Callhooks shouldn't revert, but if they do:
             // we revert, so that the retryable ticket can be re-attempted
```
tmigone, line 245:
```suggestion
```

This no longer applies, at least the explicit comment

contracts/gateway/ICallhookReceiver.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+/**
+ * @title Interface for contracts that can receive callhooks through the Arbitrum GRT bridge
+ * @dev Any contract that can receive a callhook on L2, send through the bridge from L1, must
```
tmigone, line 5:
```suggestion
 * @dev Any contract that can receive a callhook on L2, sent through the bridge from L1, must
```

tmigone (commented):
Nice! 

It might be overkill but how about adding another mock contract that doesn't implement the callhook interface and then adding a test to verify the callhook execution fails?

pcarranzav:
Good idea on the additional test @tmigone - we can use any contract so I just used RewardsManager rather than adding a new mock

tmigone (approved):

PR #723, opened 2022-09-29
chore: clean and rebuild project when changing branches

tmigone:
This ensures we have clean dependencies when switching branches.  


abarmat (commented):
Interesting idea, but might inadvertently trigger pkg installing which takes a considerable amount of time. I tried it and works great but not sure if we want to make it a permanent thing.

PR #730, opened 2022-10-04
test: fix race condition in BridgeEscrow tests

pcarranzav:


tmigone (approved):
Looks good! 

Someday I'd like to use `chai-as-promised` which allows for expressions that look/read better imo.

PR #700, opened 2022-09-01
L2: linear rewards, and minting in L2

pcarranzav:

PR #699, opened 2022-09-01
feat: add Arbitrum GRT bridge (rebased, bridge only)

pcarranzav:
Deployment commands in the CLI are also updated to include an L2 deployment.
Configuration and address book entries for Arbitrum are added as well.

This is a rebase of #552 and related PRs, including _only_ the bridge (no changes to staking or rewards).

The changes in this PR correspond to [GIP-0031](https://forum.thegraph.com/t/gip-0031-arbitrum-grt-bridge/3305)

PR #733, opened 2022-10-20
fix: prevent unpausing gateways if they're not configured (OZ M-01 from #700)

pcarranzav:
Based on a finding from OZ's audit of #700, but targeting #699 as the change is relevant to that branch too.

### DIFF
--- a/contracts/gateway/GraphTokenGateway.sol
+++ b/contracts/gateway/GraphTokenGateway.sol
@@ -45,6 +45,9 @@ abstract contract GraphTokenGateway is GraphUpgradeable, Pausable, Managed, ITok
      * @param _newPaused New value for the pause state (true means the transfers will be paused)
      */
     function setPaused(bool _newPaused) external onlyGovernorOrGuardian {
+        if (!_newPaused) {
+            _checksBeforeUnpause();
+        }
         _setPaused(_newPaused);
     }
 
@@ -54,4 +57,10 @@ abstract contract GraphTokenGateway is GraphUpgradeable, Pausable, Managed, ITok
     function paused() external view returns (bool) {
         return _paused;
     }
+
+    /**
+     * @dev Runs state validation before unpausing, reverts if
+     * something is not set properly
+     */
+    function _checksBeforeUnpause() internal view virtual;
 }
--- a/tasks/deployment/unpause.ts
+++ b/tasks/deployment/unpause.ts
@@ -2,17 +2,28 @@ import { task } from 'hardhat/config'
 import { cliOpts } from '../../cli/defaults'
 import GraphChain from '../../gre/helpers/network'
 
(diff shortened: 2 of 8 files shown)


abarmat (approved):

contracts/gateway/GraphTokenGateway.sol
```diff
@@ -54,4 +57,10 @@ abstract contract GraphTokenGateway is GraphUpgradeable, Pausable, Managed, ITok
     function paused() external view returns (bool) {
         return _paused;
     }
+
+    /**
+     * @dev Runs state validation before unpausing, reverts if
+     * something is not set properly
+     */
+    function _checksBeforeUnpause() internal view virtual;
```
tmigone, line 65: How about turning this function into a modifier? 

```
    function setPaused(bool _newPaused) external onlyGovernorOrGuardian onlyValidState(_newPaused) {
        if (!_newPaused) {
            _checksBeforeUnpause();
        }
        _setPaused(_newPaused);
    }
```

tasks/deployment/unpause.ts
```diff
@@ -2,18 +2,29 @@ import { task } from 'hardhat/config'
...
+    const { Controller } = graph.contracts
 
     console.log('> Unpausing protocol')
     const tx = await Controller.connect(governor).setPaused(false)
     await tx.wait()
 
+    console.log('Done!')
+  })
+
+task('migrate:unpause-bridge', 'Unpause bridge')
```
tmigone, line 20: Same change as https://github.com/graphprotocol/contracts/pull/717/files#diff-729e91d21c664fed317c8bccc772f1c5cce6a8a4d943d700359ad371302aef55 

which is pending merge after audit. I'd still make the change here and later we solve the conflict whenever it shows up.

tmigone (approved):

contracts/gateway/GraphTokenGateway.sol
tmigone: hmm on second thought it looks kind of weird since the new paused state shouldn't be related to the state validity

tasks/deployment/unpause.ts
pcarranzav: Oooh oops, sorry I missed that. But yeah, let's keep the change here so that we have e2e passing here.

contracts/gateway/GraphTokenGateway.sol
pcarranzav: Yeah, using a modifier could work, but it did feel a bit weird so I think I prefer it as a function... I don't feel strongly about it either way

tmigone: Yeah i think it looks weird so lets keep it as is.

abarmat (approved):

PR #735, opened 2022-10-21
fix: remove unnecessary L2 gas check (OZ N-01 from #700)

pcarranzav:


tmigone (approved):

tmigone (approved):

abarmat (approved):
I understand that we are removing this because of the validation performed by the core post-nitro contracts, is that right?

pcarranzav:
> I understand that we are removing this because of the validation performed by the core post-nitro contracts, is that right?

Exactly, see https://github.com/OffchainLabs/nitro/blob/master/contracts/src/bridge/Inbox.sol#L413

PR #736, opened 2022-10-21
fix: use strict < for block in L2 mint allowance (OZ N-02)

pcarranzav:


tmigone (approved):
This one is a great catch.

PR #739, opened 2022-10-21
fix: add storage gaps to GraphTokenGateway and GraphTokenUpgradeable (C4 M-244)

pcarranzav:
Resolves https://github.com/code-423n4/2022-10-thegraph-findings/issues/244

### DIFF
--- a/contracts/l2/token/GraphTokenUpgradeable.sol
+++ b/contracts/l2/token/GraphTokenUpgradeable.sol
@@ -50,6 +50,8 @@ contract GraphTokenUpgradeable is GraphUpgradeable, Governed, ERC20BurnableUpgra
     bytes32 private DOMAIN_SEPARATOR;
     mapping(address => bool) private _minters;
     mapping(address => uint256) public nonces;
+    // Storage gap added in case we need to add state variables to this contract
+    uint256[47] private __gap;
 
     // -- Events --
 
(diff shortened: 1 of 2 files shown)


tmigone (approved):

contracts/l2/token/GraphTokenUpgradeable.sol
```diff
@@ -50,6 +50,8 @@ contract GraphTokenUpgradeable is GraphUpgradeable, Governed, ERC20BurnableUpgra
     bytes32 private DOMAIN_SEPARATOR;
     mapping(address => bool) private _minters;
     mapping(address => uint256) public nonces;
+    // Storage gap added in case we need to add state variables to this contract
+    uint256[47] private __gap;
```
abarmat, line 54: We could just keep the gap at 50 until an upgrade happens. But it is nothing critical

abarmat (approved):

contracts/l2/token/GraphTokenUpgradeable.sol
pcarranzav: yeah, I chose 47 so that for both contracts we use a standard number of total used slots, i.e. 50

trust1995:
Consider futureproofing and lining up the Managed.sol to be a 50 slot contract as well.

pcarranzav:
> Consider futureproofing and lining up the Managed.sol to be a 50 slot contract as well.
Same goes for Pausable, Governed contracts.
Without adding a [50] __gap slot for each base contract, that component will not be upgradeable.

It's a good call, but those other contracts are used by several protocol contracts that are already deployed, so adding a storage gap would break updates in all of those. Luckily the scope for Managed, Pausable and Governed is limited and the behavior is well-defined, so we don't expect to need to add any storage variables to those.

PR #742, opened 2022-10-24
c4: gas optimization fixes

tmigone:
See https://www.notion.so/edgeandnode/Code4rena-gas-report-notes-3cae3d2fc0304724aec0311b791fa1a7 for details.

### DIFF
--- a/contracts/gateway/BridgeEscrow.sol
+++ b/contracts/gateway/BridgeEscrow.sol
@@ -36,7 +36,6 @@ contract BridgeEscrow is Initializable, GraphUpgradeable, Managed {
      * @param _spender Address of the spender that will be revoked
      */
     function revokeAll(address _spender) external onlyGovernor {
-        IGraphToken grt = graphToken();
-        grt.decreaseAllowance(_spender, grt.allowance(address(this), _spender));
+        graphToken().approve(_spender, 0);
     }
 }
--- a/contracts/gateway/L1GraphTokenGateway.sol
+++ b/contracts/gateway/L1GraphTokenGateway.sol
@@ -143,7 +143,8 @@ contract L1GraphTokenGateway is Initializable, GraphTokenGateway, L1ArbitrumMess
      * @param _escrow Address of the BridgeEscrow
      */
     function setEscrowAddress(address _escrow) external onlyGovernor {
-        require(_escrow != address(0) && Address.isContract(_escrow), "INVALID_ESCROW");
+        require(_escrow != address(0), "INVALID_ESCROW");
+        require(Address.isContract(_escrow), "MUST_BE_CONTRACT");
         escrow = _escrow;
         emit EscrowAddressSet(_escrow);
     }
--- a/contracts/gateway/L1GraphTokenGateway.sol
+++ b/contracts/gateway/L1GraphTokenGateway.sol
@@ -285,7 +285,7 @@ contract L1GraphTokenGateway is Initializable, GraphTokenGateway, L1ArbitrumMess
      * @return Base ether value required to keep retryable ticket alive
      * @return Additional data sent to L2
      */
-    function parseOutboundData(bytes memory _data)
+    function parseOutboundData(bytes calldata _data)
         private
         view
         returns (
--- a/contracts/gateway/L1GraphTokenGateway.sol
+++ b/contracts/gateway/L1GraphTokenGateway.sol
@@ -203,7 +203,7 @@ contract L1GraphTokenGateway is Initializable, GraphTokenGateway, L1ArbitrumMess
     ) external payable override notPaused returns (bytes memory) {
         IGraphToken token = graphToken();
         require(_l1Token == address(token), "TOKEN_NOT_GRT");
-        require(_amount > 0, "INVALID_ZERO_AMOUNT");
+        require(_amount != 0, "INVALID_ZERO_AMOUNT");
         require(_to != address(0), "INVALID_DESTINATION");
 
         // nested scopes to avoid stack too deep errors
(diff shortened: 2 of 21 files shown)


contracts/governance/Governed.sol
```diff
@@ -51,15 +51,16 @@ contract Governed {
      * This function must called by the pending governor.
      */
     function acceptOwnership() external {
+        address _pendingGovernor = pendingGovernor;
```
pcarranzav, line 54: I think maybe this line could be `address oldPendingGovernor = pendingGovernor` and then you don't need to define it below? and just reuse this for the require checks

contracts/governance/Managed.sol
```diff
@@ -28,6 +28,14 @@ contract Managed {
     mapping(bytes32 => address) private addressCache;
     uint256[10] private __gap;
 
+    // Immutables
+    bytes32 immutable CURATION = keccak256("Curation");
```
pcarranzav, line 32: Might be better to define visibility? Probably private?

pcarranzav (commented):
Looking good, just a couple details

pcarranzav:
Ah sorry I jumped the gun again @tmigone 😆

tmigone:
To be fair i didn't set it to draft :P

contracts/governance/Governed.sol
tmigone: ah good catch.

pcarranzav (approved):
LGTM 💯 

Would be nice to get an extra pair of eyes on the assembly changes in GraphProxy though 😨

tmigone:
yes, that's so scary that I wondered wether or not to include them 

cc @abarmat

contracts/gateway/BridgeEscrow.sol
```diff
@@ -36,7 +36,6 @@ contract BridgeEscrow is Initializable, GraphUpgradeable, Managed {
      * @param _spender Address of the spender that will be revoked
      */
     function revokeAll(address _spender) external onlyGovernor {
```
abarmat, line 38: @tmigone do we have a test for calling revokeAll?

tmigone: yes we have one here: https://github.com/graphprotocol/contracts/blob/0f553c9d1ad68d626a8f1cbc44e9c24c32f45400/test/gateway/bridgeEscrow.test.ts#L67-L75

contracts/governance/Managed.sol
```diff
@@ -28,6 +28,14 @@ contract Managed {
     mapping(bytes32 => address) private addressCache;
     uint256[10] private __gap;
 
+    // Immutables
```
abarmat, line 31: Is it better to be immutable than const? as there are no parametrization happening

tmigone: I think in this case it's the same as we are using `bytes32` to store the result (https://docs.soliditylang.org/en/v0.8.11/contracts.html#constant-and-immutable-state-variables). 

I went with `immutable` because the reports suggested it, but should be the same as far as I understand.

contracts/l2/gateway/L2GraphTokenGateway.sol
```diff
@@ -238,8 +238,7 @@ contract L2GraphTokenGateway is GraphTokenGateway, L2ArbitrumMessenger, Reentran
         if (_data.length > 0) {
             bytes memory callhookData;
             {
-                bytes memory gatewayData;
-                (gatewayData, callhookData) = abi.decode(_data, (bytes, bytes));
+                (, callhookData) = abi.decode(_data, (bytes, bytes));
```
abarmat, line 241: @pcarranzav do you remember the content of gatewayData? where was this coming from?

pcarranzav: Yeah that data is just sent by the Arbitrum router when creating the default bridge + L2 tokens, so we don't need it but we keep the format for compatibility

pcarranzav: So in more detail:

The data is decoded here: https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ArbitrumGateway.sol#L235-L255

And then used here: https://github.com/OffchainLabs/arbitrum/blob/master/packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/gateway/L2ERC20Gateway.sol#L79-L102

Which then initializes the L2 token here: https://github.com/OffchainLabs/arbitrum/blob/288488a64d0510d7eb7f9f83b346fa00feb1bb02/packages/arb-bridge-peripherals/contracts/tokenbridge/arbitrum/StandardArbERC20.sol#L44-L80

abarmat (approved):

contracts/gateway/L1GraphTokenGateway.sol
```diff
@@ -203,8 +203,8 @@ contract L1GraphTokenGateway is Initializable, GraphTokenGateway, L1ArbitrumMess
         bytes calldata _data
     ) external payable override notPaused returns (bytes memory) {
         IGraphToken token = graphToken();
-        require(_l1Token == address(token), "TOKEN_NOT_GRT");
         require(_amount != 0, "INVALID_ZERO_AMOUNT");
+        require(_l1Token == address(token), "TOKEN_NOT_GRT");
```
trust1995, line 207: Isn't it better to delay this check after the next line as well?

pcarranzav: Good catch, noted to be added in a future PR

PR #744, opened 2022-10-25
feat: avoid zero grt qfee collected being added to rebate pool

juanmardefago:
Adds some checks to avoid adding zero qfee collected allocations to the rebate pools, to avoid diluting the effective allocation stake of actual participants that provided qfees to the rebate pool.

Should improve burn rates on the rebate pools significantly

### DIFF
--- a/contracts/staking/Staking.sol
+++ b/contracts/staking/Staking.sol
@@ -1010,9 +1010,21 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking, Multicall {
             // When allocation is closed redirect funds to the rebate pool
             // This way we can keep collecting tokens even after the allocation is closed and
             // before it gets to the finalized state.
+            // If the allocation didn't collect anything before, also add
+            // effectiveAllocation to the rebatePool
             if (allocState == AllocationState.Closed) {
                 Rebates.Pool storage rebatePool = rebates[alloc.closedAtEpoch];
-                rebatePool.fees = rebatePool.fees.add(queryFees);
+                if (queryFees == alloc.collectedFees) {
+                    // Account collected fees and effective allocation in rebate pool for the epoch
+                    // Might need to initialize the rebatePool since it could not have been
+                    // initialized during close since collectedFees was 0
+                    if (!rebatePool.exists()) {
+                        rebatePool.init(alphaNumerator, alphaDenominator);
+                    }
+                    rebatePool.addToPool(alloc.collectedFees, alloc.effectiveAllocation);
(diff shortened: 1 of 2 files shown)


pcarranzav (approved):
LGTM, just one gas-related comment

(and I suppose we should get this through an audit)

contracts/staking/Staking.sol
```diff
@@ -1010,9 +1010,21 @@ contract Staking is StakingV2Storage, GraphUpgradeable, IStaking, Multicall {
             // When allocation is closed redirect funds to the rebate pool
             // This way we can keep collecting tokens even after the allocation is closed and
             // before it gets to the finalized state.
+            // If the allocation didn't collect anything before, also add
+            // effectiveAllocation to the rebatePool
             if (allocState == AllocationState.Closed) {
                 Rebates.Pool storage rebatePool = rebates[alloc.closedAtEpoch];
-                rebatePool.fees = rebatePool.fees.add(queryFees);
+                if (queryFees == alloc.collectedFees) {
```
abarmat, line 1017: Couldn't this part just do `if (!rebatePool.exists())` instead of `if (queryFees == alloc.collectedFees)` ? not sure we need the extra if

I'm thinking if we can wrap this functionality into addToPool

juanmardefago: I added the if in case `rebatePool.exists()` was quite gas intensive, but I assume that yes, it could be directly checked, the only thing is that the qfee check is also important to ratify whether we should add only the qfees part, or both the qfees and effective allocation (mirroring the behaviour of allocation closed)

abarmat: `.exists()` is one SLOAD and for the `alloc.collectedFees` we are using one too. I'll think if I can find a more compact way to do it, but the logic makes sense.

Another thing that I want to check is that nothing goes wrong with the `unclaimedAllocationsCount` in the zero fee cases

pcarranzav: We could use a local variable above (line 1008):

```
            // Add funds to the allocation
            allocCollectedFees = alloc.collectedFees.add(queryFees);
            alloc.collectedFees = allocCollectedFees
```

and then use `allocCollectedFees == queryFees` - that should save us the SLOAD?

pcarranzav: (unless the compiler / optimizer do this already, probably worth checking with a gas test)

pcarranzav:
Closing this one now that exponential rebates have been deployed, so there's no pool anymore

PR #746, opened 2022-11-09
feat: bump hardhat verify to support multiple chains

abarmat:


abarmat:
e2e tests are failing, I wonder if related to reading the config with no metadata

pcarranzav:
Seems like it, afaict the e2e uses a lot of `getItemValue` calls for the config, that uses the YAML.Document object that is returned only when retainMetadata is set to true. Was there a reason why you needed to set it to false?

gre/gre.ts
```diff
@@ -178,7 +178,7 @@ function buildGraphNetworkEnvironment(
     chainId: chainId,
     provider: provider,
     addressBook: lazyObject(() => getAddressBook(addressBookPath, chainId.toString())),
-    graphConfig: lazyObject(() => readConfig(graphConfigPath, true)),
+    graphConfig: lazyObject(() => readConfig(graphConfigPath, false)),
```
pcarranzav, line 181: I think this is what's breaking the e2e

PR #747, opened 2022-11-11
feat: scratch testnet 2 with GNS migration helpers (for testing only)

pcarranzav:


pcarranzav:
This will be replaced with a scratch-2 with the linear rewards PR, and a scratch-3 with updated migration helpers.

PR #748, opened 2022-11-14
fix(L1GraphTokenGateway): add a counterpartGateway getter for Router compatibility

pcarranzav:


abarmat (approved):
💯

pcarranzav:
I'll go ahead and merge, note this didn't make it into the latest audits but is a safe change as it only adds a getter (view) function.

PR #750, opened 2022-11-21
fix(L2GraphTokenGateway): burn without allowance for Router / Arbitrum UI compatibility

pcarranzav:
Looks like the Arbitrum router and bridge UI expects the L2 gateway to be able to burn tokens without requiring allowance.


pcarranzav:
This can be handled in the Arbitrum UI without need for this PR, so closing for now

PR #755, opened 2022-11-30
feat: added tasks to list bridge txs

tmigone:
Screen Shot 2022-11-30 at 00 01 28


pcarranzav (approved):
Nice!

PR #758, opened 2022-12-01
chore: manual action for verifying deployments

spalladino:
Adds a manual github action for verifying deployments. The inputs to the action are the list of contracts to verify and their network. From that, the deployment addresses are picked up from `addresses.json`.

Screenshot from 2022-12-01 12-43-13

The verify job will use [Defender's API](https://docs.openzeppelin.com/defender/admin#verifying_any_contract) to check that the compiled artifacts at that branch match the ones deployed. An example run can be seen [here](https://github.com/spalladino/graphprotocol-contracts/actions/runs/3594057672).

Screenshot from 2022-12-01 12-42-56

These results can also be manually checked on Defender after the verification was run.

Screenshot from 2022-12-01 12-42-26

This workflow requires defining the repository secrets:
- `DEFENDER_API_KEY`
- `DEFENDER_API_SECRET`
- `INFURA_KEY`

### DIFF
--- a/tasks/verify/defender.ts
+++ b/tasks/verify/defender.ts
@@ -1,0 +1,98 @@
+import { execSync } from 'child_process'
+import { task } from 'hardhat/config'
+import { HardhatRuntimeEnvironment as HRE } from 'hardhat/types'
+import { constants } from 'ethers'
+import { appendFileSync } from 'fs'
+import type { VerificationResponse } from '@openzeppelin/hardhat-defender/dist/verify-deployment'
+
+async function main(args: { referenceUrl?: string; contracts: string[] }, hre: HRE) {
+  const { referenceUrl, contracts } = args
+  const { defender, network, graph } = hre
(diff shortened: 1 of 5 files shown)


tasks/verify/defender.ts
```diff
@@ -0,0 +1,98 @@
...
+    throw new Error(
+      `Some verifications failed:\n${errs
+        .map(([name, err]) => ` ${name}: ${err.message}`)
+        .join('\n')}`,
+    )
+  }
+}
+
+function etherscanLink(network: string, address: string): string {
+  switch (network) {
```
pcarranzav, line 70: nit: it would be good to include cases for goerli and arbitrum-goerli, since those are valid options for the workflow

pcarranzav (approved):
This looks great! Left a small comment but otherwise LGTM

pcarranzav (approved):

PR #766, opened 2022-12-16
feat: changed dispute manager to not delete bad actors disputes on re…

MercuricChloride:
…solution

Related to GIP-0041. This shouldn't cause any issues regarding functionality of the dispute manager, but not deleted malicious disputeIDs allows the subgraph bridge to not allow accepted dispute attestations to be used on chain.

FORUM POST: https://forum.thegraph.com/t/gip-0041-store-dispute-status-on-resolution/3759

Any feedback is welcome. 

Thank you!

### DIFF
--- a/contracts/disputes/DisputeManager.sol
+++ b/contracts/disputes/DisputeManager.sol
@@ -563,7 +563,7 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
      * @param _disputeID ID of the dispute to be accepted
      */
     function acceptDispute(bytes32 _disputeID) external override onlyArbitrator {
-        Dispute memory dispute = _resolveDispute(_disputeID);
+        Dispute memory dispute = disputes[_disputeID];
 
         // Slash
         (, uint256 tokensToReward) = _slashIndexer(


pcarranzav (changes requested):
Hi @MercuricChloride! Thanks for sending this in!

I'm afraid this change would make the acceptDispute function vulnerable to a malicious arbitrator calling it repeatedly and slashing the indexer multiple times. (Arbitrators are generally trusted, but the contract should still not allow doing more than they're supposed to).

So I think for this change to work, we'd need the additional changes described in the linked GIP, i.e. adding a DisputeStatus and using it to validate that the dispute is pending before accepting it.

MercuricChloride:
> Hi @MercuricChloride! Thanks for sending this in!
> 
> I'm afraid this change would make the acceptDispute function vulnerable to a malicious arbitrator calling it repeatedly and slashing the indexer multiple times. (Arbitrators are generally trusted, but the contract should still not allow doing more than they're supposed to).
> 
> So I think for this change to work, we'd need the additional changes described in the linked GIP, i.e. adding a DisputeStatus and using it to validate that the dispute is pending before accepting it.

Sure that makes sense, i'll make some changes and submit them here shortly.

MercuricChloride:
Just changed the code to include a new mapping `disputeStatus` which goes from disputeID -> DisputeStatus enum. My reasoning for having this be a separate mapping and not part of the Dispute struct, is to modify the existing structure of the contracts as little as possible. But maybe we should just refactor how that works and store it within the struct and not delete the Dispute struct in storage, just update the enum value within the _resolveDispute functions. Let me know what you think.

contracts/disputes/IDisputeManager.sol
```diff
@@ -12,6 +12,13 @@ interface IDisputeManager {
         QueryDispute
     }
 
+    enum DisputeStatus {
+        Accepted,
```
pcarranzav, line 16: This will make non-existent disputes show up as Accepted, because it will be the default status (0). I'd suggest adding a `Null` status or similar as first value for the enum.

pcarranzav (commented):
Thanks for incorporating the feedback @MercuricChloride.

Yeah, I actually think it's better if we go the full way and modify the Dispute struct to include the status. We can then change the check if a dispute exists by looking for a Pending status, and instead of deleting it when it's accepted/drawn/rejected we set the right status.

This means a couple more changes probably, so feel free to reach out to me on Discord if  it helps to go over the code together.

contracts/disputes/IDisputeManager.sol
MercuricChloride: Ok cool sounds like a plan! I'll modify those here in a bit. I'll also reach out on discord.

contracts/disputes/DisputeManager.sol
```diff
@@ -611,14 +639,24 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
...
+        // store dispute status
+        dispute.status = IDisputeManager.DisputeStatus.Drawn;
 
         // Return deposit to the fisherman
         TokenUtils.pushTokens(graphToken(), dispute.fisherman, dispute.deposit);
 
         // Resolve the conflicting dispute if any
-        _resolveDisputeInConflict(dispute);
+        if (_isDisputeInConflict(dispute)) {
+            drawDispute(dispute.relatedDisputeID);
```
pcarranzav, line 658: This will end up pushing the tokens twice. I think you can leave this part as it was before, and simply change _resolveDisputeInConflict to mark the conflicting dispute as Drawn

contracts/disputes/DisputeManager.sol
```diff
@@ -628,39 +666,42 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
...
      * @dev Returns whether the dispute is for a conflicting attestation or not.
      * @param _dispute Dispute
      * @return True conflicting attestation dispute
      */
-    function _isDisputeInConflict(Dispute memory _dispute) private pure returns (bool) {
-        return _dispute.relatedDisputeID != 0;
+    function _isDisputeInConflict(Dispute memory _dispute) private view returns (bool) {
+        bytes32 relatedID = _dispute.relatedDisputeID;
+        return
+            relatedID != 0 && disputes[relatedID].status == IDisputeManager.DisputeStatus.Pending;
```
pcarranzav, line 688: Not sure why this was added? I think as long as the relatedDisputeID is set it should be sufficient to know there's a conflicting dispute?

contracts/disputes/DisputeManager.sol
```diff
@@ -154,6 +154,15 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
         _;
     }
 
+    modifier onlyPendingDispute(bytes32 _disputeID) {
+        require(isDisputeCreated(_disputeID), "Dispute does not exist");
```
pcarranzav, line 158: Seems like this line can be removed, as the check for Pending will also mean checking for not Null.

contracts/disputes/DisputeManager.sol
```diff
@@ -628,24 +664,26 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
      * @param _disputeID ID of the dispute to resolve
      * @return Dispute
      */
-    function _resolveDispute(bytes32 _disputeID) private returns (Dispute memory) {
-        require(isDisputeCreated(_disputeID), "Dispute does not exist");
+    // function _resolveDispute(bytes32 _disputeID) private returns (Dispute memory) {
```
pcarranzav, line 667: nit: let's remove commented code for safety and clarity please

contracts/disputes/DisputeManager.sol
pcarranzav: Ah, so now I see this is for the `_isDisputeInConflict` check to return false when rejecting the related dispute. This makes sense, but it would be good to add a comment here to explain it.

contracts/disputes/DisputeManager.sol
```diff
@@ -656,7 +694,8 @@ contract DisputeManager is DisputeManagerV1Storage, GraphUpgradeable, IDisputeMa
     function _resolveDisputeInConflict(Dispute memory _dispute) private returns (bool) {
```
pcarranzav, line 694: Since this is now only used for Draws, it'd be good to rename the function to clarify it, e.g. `_resolveDrawnDisputeInConflict` or `_drawDisputeInConflict`.

test/disputes/query.test.ts
```diff
@@ -516,7 +517,7 @@ describe('DisputeManager:Query', async () => {
       await disputeManager.connect(arbitrator.signer).acceptDispute(dID1)
       // Check
       const relatedDispute = await disputeManager.disputes(dID2)
-      expect(relatedDispute.indexer).eq(AddressZero)
+      expect(relatedDispute.status).not.eq(3) // 3 = DisputeStatus.Pending
```
pcarranzav, line 520: It'd be better to explicitly check for the status we expect here, rather than the one we don't. It'd also be nice to add similar checks in other tests here to have better coverage and ensure the status changes work as expected for other cases.

pcarranzav (commented):
A few more comments but looking better 👍

pcarranzav:
Important conclusion from the audit to include in the deployment plan: ensure all active disputes are resolved before executing this upgrade.

pcarranzav (approved):

PR #783, opened 2023-01-16
feat: make bridge tasks prettier and display more info for L2 to L1

tmigone:
### Changes
- bridge tasks
   - Format GRT amounts to 2 decimal places, makes is easier to read/scan task output
   - Show granular information for withdrawal amounts
   - Show L1 and L2 tx's related to a deposit/withdrawal (where possible, in L1 to L2 arb node rate limiting prevents from getting retryable attempts).

Screen Shot 2023-01-18 at 13 28 05

### DIFF
--- a/tasks/bridge/deposits.ts
+++ b/tasks/bridge/deposits.ts
@@ -37,6 +37,11 @@ task(TASK_BRIDGE_DEPOSITS, 'List deposits initiated on L1GraphTokenGateway')
       `Tracking 'DepositFinalized' events on L2GraphTokenGateway (${l2Gateway.address}) from block ${l2StartBlock} onwards`,
     )
 
+    const amount = await (
+      await l1Gateway.queryFilter(l1Gateway.filters.DepositInitiated(), l1StartBlock, l1EndBlock)
+    ).length
+    console.log(`Found ${amount} deposits on L1`)
+
     const depositInitiatedEvents = await Promise.all(
       (
         await l1Gateway.queryFilter(l1Gateway.filters.DepositInitiated(), l1StartBlock, l1EndBlock)
@@ -76,7 +81,6 @@ task(TASK_BRIDGE_DEPOSITS, 'List deposits initiated on L1GraphTokenGateway')
 
     printEvents(depositInitiatedEvents)
     console.timeEnd('runtime')
-    console.timeLog('runtime')
   })
 
 function printEvents(events: any[]) {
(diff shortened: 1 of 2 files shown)


pcarranzav (dismissed):
LGTM

tasks/bridge/deposits.ts
```diff
@@ -16,66 +16,90 @@ task(TASK_BRIDGE_DEPOSITS, 'List deposits initiated on L1GraphTokenGateway')
...
 
     console.log(
-      'L1 to L2 message status reference: 🚧 = not yet created, ❌ = creation failed, ⚠️  = funds deposited on L2, ✅ = redeemed, ⌛ = expired',
+      '\nL1 to L2 message status reference: 🚧 = not yet created, ❌ = creation failed, ⚠️  = funds deposited on L2, ✅ = redeemed, ⌛ = expired',
     )
 
-    printEvents(events)
+    printEvents(depositInitiatedEvents)
+    console.timeEnd('runtime')
+    console.timeLog('runtime')
```
pcarranzav, line 79: nit: TIL about `console.time` - but I tried locally and calling timeLog after timeEnd doesn't seem to work?

```
Welcome to Node.js v16.18.0.
Type ".help" for more information.
> console.time('foo')
undefined
> console.timeEnd('foo')
foo: 5.500s
undefined
> console.timeLog('foo')
undefined
> (node:4083) Warning: No such label 'foo' for console.timeLog()
(Use `node --trace-warnings ...` to show where the warning was created)

> 
```

pcarranzav (approved):
LGTM

tasks/bridge/deposits.ts
tmigone: Ah good catch. You can use `timeLog` to log partial times before ending the timer with `timeEnd`. Don't remember why this one was here, probably a mistake.

tasks/bridge/deposits.ts
```diff
@@ -37,6 +37,11 @@ task(TASK_BRIDGE_DEPOSITS, 'List deposits initiated on L1GraphTokenGateway')
       `Tracking 'DepositFinalized' events on L2GraphTokenGateway (${l2Gateway.address}) from block ${l2StartBlock} onwards`,
     )
 
+    const amount = await (
```
pcarranzav, line 40: did you add this additional query on purpose? if so, it might be better to store the result and use it for depositInitiatedEvents below?

But also, isn't it this redundant with what's printed on line 73?

pcarranzav (approved):
(I left that small comment above but this still LGTM)

tasks/bridge/deposits.ts
tmigone: Fixed!

PR #797, opened 2023-03-10
feat(rebates): introduce exponential rebates

tmigone:
This PR adds exponential rebates for new rebate pools. 

A few notes about the implementation:
- Does not remove support for Cobb Douglas rebate pools
- Any pre-existing rebate pool will continue to be governed by the CD rebates formula
- Any new created rebate pools will be governed by the exponential rebates formula
- Removes `claimMany` and `closeAllocationMany` from the `Staking` contract to be within max contract size limit. Indexer stack needs to be updated accordingly to use multicall instead of these functions.

This PR is the initial implementation for GIP-0051, discussion post here: https://forum.thegraph.com/t/gip-0051-exponential-query-fee-rebates-for-indexers/4162

Note that the GIP has diverged since this was implemented so a follow up PR will bring further changes (notably removal of Cob Douglas and rebate pools).


contracts/staking/libs/Rebates.sol
```diff
@@ -88,15 +103,29 @@ library Rebates {
...
+                    _indexerFees,
+                    _indexerEffectiveAllocatedStake,
+                    pool.alphaNumerator,
+                    pool.alphaDenominator,
+                    pool.lambdaNumerator,
+                    pool.lambdaDenominator
+                );
+            } else {
+                // (Deprecated) Cobb-Douglas rebates
+                // Keeping this logic to allow existing open pools to be closed
```
abarmat, line 118: Maybe we can find a different upgrade path so that we don't need this code. What if use the new formula for old rebates?

tmigone: I thought about that, but is it _fair_ to indexers? Going to ask Justin/Howard about this.

tmigone:
This was superseded by https://github.com/graphprotocol/contracts/pull/818, which includes the changes in here plus rebate pool removal.

PR #824, opened 2023-05-04
feat(rebates): change query fee rebates mechanism to use exponential formula [rebased]

tmigone:
__Note__ this PR is the same as https://github.com/graphprotocol/contracts/pull/818 but with all the migratoor (now transfer tools ™️) changes rebased. 

### Motivation

This PR implements the required changes for [GIP-0051](https://forum.thegraph.com/t/gip-0051-exponential-query-fee-rebates-for-indexers/4162). For a detailed rationale and changes involved please read the document.

### Changes
**Staking contract changes**
- Remove `closeAllocationMany` and `claimMany` functions (unrelated to GIP)
- Replace Cob-Douglass rebates with exponential rebates
    - Remove CobDouglass library
    - Add exponential rebates library
    - Add `lambdaNumerator` and `lambdaDenominator` to staking contract storage. For `alphaNumerator` and `alphaDenominator` we will re-use the storage variables currently in use for the Cobb-Douglas formula.
- Update `collect()` function to use exponential rebates
    - Remove `claim()` function. Combine it with `collect()` in a single operation for better indexer UX
    - Calculate rebated fees using exponential rebates formula
    - Re-stake rebated fees only if `rewardsDestination` is not set for the indexer (`_restake` flag behavior on `claim()` removed)
    - Replace `AllocationCollected` with `RebatesCollected` event
    - Additionally, we updated `collect()` to allow collecting multiple vouchers for the same allocation.
    - Add `distributedRebates` to the `Allocation` struct to keep track of rebate distribution across multiple voucher collections.
- Remove rebate pools 🥳 
    - Deprecate `rebates` storage variable
    - Deprecate `channelDisputeEpochs` storage variable and remove associated setters/getters
    - Deprecate `effectiveAllocation` in `Allocation` struct
    - Remove `Claimed` and `Finalized` states from `Allocation` life cycle, affects `getAllocationState()`
    - Remove Rebates library
    - Update `closeAllocation()` function to remove rebate pools logic.

Some notes about the changes:
- The new `collect()` function can collect multiple vouchers for the same allocation regardless of the gateway that signed them. Currently there is a single gateway but this implementation should be valid for a permissioned multi-gateway scenario, which is something we want to achieve in the near future.
- Another thing to note regarding multiple collect calls on a single allocation is that currently the `AllocationExchange` will still limit the amount of vouchers per allocation to 1. Again this is likely to change in the future and this implementation should account for that.
- Regarding the `claim()` function merge into `collect()`, we keep the same name and function signature to avoid changing and upgrading the `AllocationExchange`. This also means no changes are required in the indexer stack.


test/staking/delegation.test.ts
```diff
@@ -593,68 +594,6 @@ describe('Staking::Delegation', () => {
       expect(await staking.getIndexerCapacity(indexer.address)).eq(0)
     })
 
-    it('should send delegation cut of query fees to delegation pool', async function () {
```
pcarranzav, line 596: I'm curious why this test was removed, is this behavior not the same anymore?

pcarranzav (approved):

test/staking/delegation.test.ts
tmigone: It is. I moved this to `allocation.test.ts`, the function `shouldCollect` tests for the same things but in a lot more scenarios.

contracts/governance/Managed.sol
```diff
@@ -8,6 +8,7 @@ import { ICuration } from "../curation/ICuration.sol";
 import { IEpochManager } from "../epochs/IEpochManager.sol";
 import { IRewardsManager } from "../rewards/IRewardsManager.sol";
 import { IStaking } from "../staking/IStaking.sol";
+import { IStakingBase } from "../staking/IStakingBase.sol";
```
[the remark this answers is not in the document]
tmigone: `RewardsManager.getRewards()` is using it. I added it here since the other contract interfaces were also here.

test/staking/delegation.test.ts
pcarranzav: Aaah makes sense, missed that, thanks

PR #831, opened 2023-05-19
feat: add upgrade testing to the e2e framework

tmigone:
This PR extends the e2e framework to include contract upgrade testing via a new script. Using a local fork from mainnet or other sources it allows writing tests to validate contract upgrades, state transitions, etc.

The script will:
- Run a hardhat node forking from an RPC (mainnet or any network)
- Runs a user defined pre-upgrade script (in case it’s useful)
- Runs a user defined pre-upgrade test script (here you can write tests like usual, using GRE, etc)
- Upgrades the specified contract in the fork using the current directory source code (this includes impersonating governance to accept upgrades)
- Runs a user defined post-upgrade script (here you might do things like contract configuration that requires the new implementation address, add it to callhook whitelist for example)
- Runs a user defined post-upgrade test script (here you can write tests to validate the deployment, maybe if the upgrade was fixing a bug you can verify it)

An example can be found in `e2e/upgrades/example`.

These changes are also included in https://github.com/graphprotocol/contracts/pull/838 which is the branch we'll likely end up merging.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ addresses-local.json
 localNetwork.json
 arbitrum-addresses-local.json
 tx-*.log
+addresses-fork.json
 
 # Keys
 .keystore
--- a/cli/network.ts
+++ b/cli/network.ts
@@ -303,6 +303,34 @@ export const deployContractAndSave = async (
   return deployResult.contract
 }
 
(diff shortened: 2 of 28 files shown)


tmigone:
Closing this PR as we'll merge these changes in #838

PR #838, opened 2023-05-24
fix: Lack of test coverage (OZ M-01)

tmigone:
This PR addresses OZ M-01 but also brings in the upgrade testing scripts from https://github.com/graphprotocol/contracts/pull/831 as they are required to perform e2e tests on forked chains to correctly evaluate upgrade state transitions.

**Notes from #831**
> This PR extends the e2e framework to include contract upgrade testing via a new script. Using a local fork from mainnet or other sources it allows writing tests to validate contract upgrades, state transitions, etc.
> 
> The script will:
> 
> - Run a hardhat node forking from an RPC (mainnet or any network)
> - Runs a user defined pre-upgrade script (in case it’s useful)
> - Runs a user defined pre-upgrade test script (here you can write tests like usual, using GRE, etc)
> - Upgrades the specified contract in the fork using the current directory source code (this includes impersonating governance to accept upgrades)
> - Runs a user defined post-upgrade script (here you might do things like contract configuration that requires the new implementation address, add it to callhook whitelist for example)
> - Runs a user defined post-upgrade test script (here you can write tests to validate the deployment, maybe if the upgrade was fixing a bug you can verify it)
> 
> An example can be found in e2e/upgrades/example.

### DIFF
--- a/cli/network.ts
+++ b/cli/network.ts
@@ -320,6 +320,34 @@ export const deployContractAndSave = async (
   return deployResult.contract
 }
 
+export const deployContractImplementationAndSave = async (
+  name: string,
+  args: Array<ContractParam>,
+  sender: Signer,
+  addressBook: AddressBook,
+): Promise<Contract> => {
+  // Deploy the contract
+  const deployResult = await deployContract(name, args, sender)
+
+  // Save address entry
+  const entry = addressBook.getEntry(name)
+  entry.implementation = {
+    address: deployResult.contract.address,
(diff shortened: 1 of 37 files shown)


cli/network.ts
```diff
@@ -320,6 +320,34 @@ export const deployContractAndSave = async (
   return deployResult.contract
 }
 
+export const deployContractImplementationAndSave = async (
```
pcarranzav, line 323: Niiiice it's great to add this

pcarranzav (approved):

PR #841, opened 2023-05-26
fix(rewards): getRewards now returns zero if allo is not active (OZ L-02)

tmigone:


contracts/rewards/RewardsManager.sol
```diff
@@ -355,6 +355,11 @@ contract RewardsManager is RewardsManagerV4Storage, GraphUpgradeable, IRewardsMa
      * @return Rewards amount for an allocation
```
abarmat, line 355: Doing staking() requires a CALL and a few SLOAD from the Controller, maybe we could store in memory before querying the state and using it then in `getAllocation`

abarmat: `IStaking staking = staking();`

PR #855, opened 2023-08-02
feat: remove asset holder allowlist (with latest main)

pcarranzav:


contracts/staking/IStakingBase.sol
```diff
@@ -93,12 +93,6 @@ interface IStakingBase is IStakingData {
         uint32 cooldownBlocks
     );
 
-    /**
-     * @dev Emitted when `caller` set `assetHolder` address as `allowed` to send funds
-     * to staking contract.
-     */
-    event AssetHolderUpdate(address indexed caller, address indexed assetHolder, bool allowed);
```
tmigone, line 100: (GitHub won't let me comment on the correct LOC as you did not modify it but...)

The event `RebateCollected` uses `assetHolder` name for the query fees provider. We probably want to change that name for clarity, something like `rebateProvider` or `provider`, `sender`, wdyt?

contracts/staking/Staking.sol
```diff
@@ -366,13 +354,17 @@ abstract contract Staking is StakingV4Storage, GraphUpgradeable, IStakingBase, M
...
 
-        // The contract caller must be an authorized asset holder
-        require(__assetHolders[msg.sender] == true, "!assetHolder");
-
         // Allocation must exist
         AllocationState allocState = _getAllocationState(_allocationID);
         require(allocState != AllocationState.Null, "!collect");
 
+        // If the query fees are zero, we don't want to revert
+        // but we also don't need to do anything, so just return
```
tmigone, line 362: This changes slightly the behavior of the function as we now are not emitting the `RebatesCollected` event when `_tokens == 0`. I can't think why this could be a problem, but wanted to make sure we consider it.

tmigone (approved):
LGTM, left some comments

pcarranzav:
Replaced by #864

PR #864, opened 2023-10-11
feat: remove asset holder allowlist, with additional fixes

pcarranzav:


contracts/staking/StakingExtension.sol
```diff
@@ -26,6 +26,8 @@ contract StakingExtension is StakingV4Storage, GraphUpgradeable, IStakingExtensi
 
     /// @dev 100% in parts per million
     uint32 private constant MAX_PPM = 1000000;
+    /// @dev Minimum delegation for the first delegator of an indexer
+    uint256 private constant MINIMUM_DELEGATION = 1e18;
```
tmigone, line 30: Too bad we need to define `MINIMUM_DELEGATION` in two different files, is it too bad if we do it in `IStakingData.sol`? Only way that both files could read it from the same place.

contracts/staking/StakingExtension.sol
```diff
@@ -555,9 +546,13 @@ contract StakingExtension is StakingV4Storage, GraphUpgradeable, IStakingExtensi
         uint256 delegatedTokens = _tokens.sub(delegationTax);
 
         // Calculate shares to issue
-        uint256 shares = (pool.tokens == 0)
-            ? delegatedTokens
-            : delegatedTokens.mul(pool.shares).div(pool.tokens);
+        uint256 shares;
+        if (pool.tokens == 0) {
+            require(_tokens >= MINIMUM_DELEGATION, "!minimum-delegation");
```
tmigone, line 551: Same as above about the comment.

contracts/l2/staking/L2Staking.sol
```diff
@@ -122,7 +125,14 @@ contract L2Staking is Staking, IL2StakingBase {
         Delegation storage delegation = pool.delegators[_delegationData.delegator];
 
         // Calculate shares to issue (without applying any delegation tax)
-        uint256 shares = (pool.tokens == 0) ? _amount : _amount.mul(pool.shares).div(pool.tokens);
+        uint256 shares;
+        if (pool.tokens == 0) {
+            if (_amount >= MINIMUM_DELEGATION) {
```
tmigone, line 130: I would add a minimal comment explaining or referencing the issue this fixes. It's not clear why we are doing this otherwise.

tmigone (approved):
Some minor nits.

contracts/staking/StakingExtension.sol
tmigone: nvm it's an interface not a contract, so we can't have variables in there.

tmigone (approved):
lgtm

contracts/l2/curation/L2Curation.sol
```diff
@@ -421,6 +421,28 @@ contract L2Curation is CurationV2Storage, GraphUpgradeable, IL2Curation {
...
 
+    /**
+     * @notice Calculate the amount of tokens that would be recovered if minting signal with
+     * the input tokens and then burning it. This can be used to compute rounding error.
+     * This function does not account for curation tax.
+     * @param _subgraphDeploymentID Subgraph deployment for which to mint signal
+     * @param _tokensIn Amount of tokens used to mint signal
+     * @return Amount of tokens that would be recovered after minting and burning signal
+     */
+    function tokensToSignalToTokensNoTax(bytes32 _subgraphDeploymentID, uint256 _tokensIn)
```
tmigone, line 432: lol @ that name

tmigone (changes requested):
looks good but we are not testing the GNS round attack no?

contracts/l2/curation/L2Curation.sol
pcarranzav: tokensToSignalToTokensNoTaxFinalFinalVersion4.docx

Maikol (approved):

PR #866, opened 2023-10-11
feat: remove/deprecate delegation parameters cooldown

pcarranzav:
GIP will be published soon

### DIFF
--- a/contracts/staking/StakingExtension.sol
+++ b/contracts/staking/StakingExtension.sol
@@ -41,18 +41,16 @@ contract StakingExtension is StakingV4Storage, GraphUpgradeable, IStakingExtensi
      * initialize() function, so it uses the same access control check to ensure it is
      * being called by the Staking implementation as part of the proxy upgrade process.
      * @param _delegationUnbondingPeriod Delegation unbonding period in blocks
-     * @param _cooldownBlocks Minimum time between changes to delegation parameters, in blocks
      * @param _delegationRatio Delegation capacity multiplier (e.g. 10 means 10x the indexer stake)
      * @param _delegationTaxPercentage Percentage of delegated tokens to burn as delegation tax, expressed in parts per million
      */
     function initialize(
         uint32 _delegationUnbondingPeriod,
-        uint32 _cooldownBlocks,
+        uint32, //_cooldownBlocks, deprecated
         uint32 _delegationRatio,
         uint32 _delegationTaxPercentage
     ) external onlyImpl {
         _setDelegationUnbondingPeriod(_delegationUnbondingPeriod);
-        _setDelegationParametersCooldown(_cooldownBlocks);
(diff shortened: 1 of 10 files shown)


contracts/staking/StakingExtension.sol
```diff
@@ -41,18 +41,16 @@ contract StakingExtension is StakingV4Storage, GraphUpgradeable, IStakingExtensi
...
      * being called by the Staking implementation as part of the proxy upgrade process.
      * @param _delegationUnbondingPeriod Delegation unbonding period in blocks
-     * @param _cooldownBlocks Minimum time between changes to delegation parameters, in blocks
      * @param _delegationRatio Delegation capacity multiplier (e.g. 10 means 10x the indexer stake)
      * @param _delegationTaxPercentage Percentage of delegated tokens to burn as delegation tax, expressed in parts per million
      */
     function initialize(
         uint32 _delegationUnbondingPeriod,
-        uint32 _cooldownBlocks,
+        uint32, //_cooldownBlocks, deprecated
```
tmigone, line 49: Do we need to keep it here? I understand doing it for `setDelegationParameters` so we don't change the interface but I think we can entirely remove it here?

contracts/staking/StakingStorage.sol
```diff
@@ -69,8 +69,8 @@ contract StakingV1Storage is Managed {
     /// then they can use up to 500 GRT from the delegated stake
     uint32 internal __delegationRatio;
 
-    /// @dev Time in blocks an indexer needs to wait to change delegation parameters
-    uint32 internal __delegationParametersCooldown;
+    /// @dev Time in blocks an indexer needs to wait to change delegation parameters (deprecated)
+    uint32 internal __DEPRECATED_delegationParametersCooldown;
```
tmigone, line 73: I think we should pick one way of signaling deprecated variables. We have `__DEPRECATED__` or `...Deprecated` all over the codebase.

tmigone (changes requested):
Some nits

tmigone:
Also we need to remove this: https://github.com/graphprotocol/contracts/blob/59240bd1543c8e33620b28a13699c2c57b3eb064/cli/commands/protocol/set.ts#L17-L20

contracts/staking/StakingExtension.sol
pcarranzav: We could, but I'd rather keep it to keep changes minimal (this is called from Staking.initialize, so we'd have to change that too)

contracts/staking/StakingStorage.sol
pcarranzav: Changed all of them to use the `__DEPRECATED_` prefix, wdyt?

Maikol (approved):

tmigone (approved):

PR #867, opened 2023-10-13
feat: sepolia testnet

Maikol:
Adding sepolia and arbitrum sepolia testnets.

### DIFF
--- a/tasks/verify/verify.ts
+++ b/tasks/verify/verify.ts
@@ -5,7 +5,7 @@ import { isFullyQualifiedName, parseFullyQualifiedName } from 'hardhat/utils/con
 import { TASK_COMPILE } from 'hardhat/builtin-tasks/task-names'
 import { loadEnv } from '../../cli/env'
 import { cliOpts } from '../../cli/defaults'
-import { getContractConfig } from '../../cli/config'
+import { getContractConfig, readConfig } from '../../cli/config'
 import { Wallet } from 'ethers'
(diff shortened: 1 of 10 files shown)


tasks/verify/verify.ts
```diff
@@ -84,10 +84,10 @@ task('verifyAll', 'Verifies all contracts on etherscan')
     }
 
     console.log(`> Verifying all contracts on chain ${chainName}[${chainId}]...`)
-    const { addressBook, graphConfig } = hre.graph({
+    const { addressBook } = hre.graph({
       addressBook: args.addressBook,
-      graphConfig: args.graphConfig,
     })
+    const graphConfig = readConfig(args.graphConfig)
```
Maikol, line 90: I had to change the way we load `graphConfig` because the task was not working for me. Let me know if it could be something wrong with my setup.

tmigone: hmm sounds like an issue with the cmd line parameters or your setup, I would revert this change and investigate what the problem was (create an issue for it?). You dont happen to have kept logs no?

tmigone (changes requested):
lgtm, left a comment on the verify task change

tasks/verify/verify.ts
Maikol: I've created a ticket for this https://github.com/graphprotocol/contracts/issues/871 thanks

tmigone (approved):

PR #872, opened 2023-10-27
chore: wip changes to integrate with sdk

tmigone:
This PR is the integration with https://github.com/graphprotocol/sdk.

It's a gigantic PR to review... so here are the sanity checks I've been using to validate the changes:
- unit tests pass `yarn test`
- e2e tests pass `L1_NETWORK=localhost yarn test:e2e`
- `hh console --network mainnet` and query some values like `await g.l2.contracts.GraphToken.totalSupply()`

### Changes
- Use `@graphprotocol/sdk` for unit tests, e2e tests and hardhat tasks.
- Unit tests now deploy the protocol using a graph config file (`graph.hardhat.yml` or `graph.arbitrum-hardhat.yml`) and an address book, and use same code/logic as e2e tests and migrate/deploy commands.
- `cli` now "deprecated". No code from `cli/` should be used anywhere. Will remove in a followup PR once we finish migrating any useful command.
- Removed `GraphGovernance` contract 
- Removed GRE from repo


package.json
```diff
@@ -1,19 +1,20 @@
 {
   "name": "@graphprotocol/contracts",
-  "version": "5.3.0",
+  "version": "5.3.2",
```
pcarranzav, line 3: nit: why not 5.3.1?

test/lib/fixtures.ts
```diff
@@ -64,181 +74,51 @@ export interface ArbitrumL1Mocks {
...
+    await contracts.GraphToken.connect(deployer).addMinter(deployer.address)
+    await contracts.Controller.connect(deployer).setPaused(false)
 
-  async loadL2(
-    deployer: Signer,
-    slasher: Signer = Wallet.createRandom() as Signer,
-    arbitrator: Signer = Wallet.createRandom() as Signer,
-  ): Promise<L2FixtureContracts> {
-    return this._loadLayer(deployer, slasher, arbitrator, true) as unknown as L2FixtureContracts
+    return contracts
```
pcarranzav, line 112: As mentioned in our call, returning `contracts as L1FixtureContracts` or `as L2FixturContracts` would solve a lot of the type wrangling in the tests, but can be addressed in a separate PR

test/governance/governed.test.ts
```diff
@@ -3,42 +3,46 @@ import hre from 'hardhat'
...
+  let me: SignerWithAddress
+  let governor: SignerWithAddress
 
   let governed: Governed
 
   beforeEach(async function () {
-    ;[me, governor] = await getAccounts()
+    ;[me, governor] = await graph.getTestAccounts()
+
+    await helpers.setAutoMine(true)
```
pcarranzav, line 22: What's the reason to add this here?

test/staking/allocation.test.ts
```diff
@@ -30,16 +31,65 @@ enum AllocationState {
   Closed,
 }
 
+const ABI_LIB_EXPONENTIAL = [
```
pcarranzav, line 34: Could this be imported from hardhat artifacts instead?

test/staking/rebate.test.ts
```diff
@@ -14,6 +16,53 @@ const toFixed = (n: number | BigNumber, precision = 12) => {
   return toFloat(n).toFixed(precision)
 }
 
+const ABI_LIB_EXPONENTIAL = [
```
pcarranzav, line 19: Same as above, could this use hardhat artifacts?

pcarranzav (approved):
Great work on this massive lift 💪 
A few small comments but LGTM.

(I'm a bit scared of the merge conflicts we'll have on the tests for some open PRs, but I guess there'll never be an ideal time, better to merge this now and fix those as they appear)

package.json
tmigone: I think 5.3.1 was claimed and published from another branch.

test/lib/fixtures.ts
tmigone: Definitely a better approach. Though the ultimate goal is to get rid of `L1FixtureContracts` and `L2FixtureContracts` in favor of `GraphNetworkContracts`. 

Created #885

test/governance/governed.test.ts
tmigone: The node would otherwise be set to not automine and be on a 13s block time (this defaults are defined in `hardhat.config.ts`. Previously the deployment/fixtures code was turning the automine on.

Now I'm not sure why the default config is to be on a 13s clock... for a local dev/test node I prefer a fast automine node, so maybe we should change that. Created https://github.com/graphprotocol/contracts/issues/886 to track this and investigate further.

test/staking/allocation.test.ts
tmigone: totally! Tracking here https://github.com/graphprotocol/contracts/issues/887

test/staking/rebate.test.ts
tmigone: https://github.com/graphprotocol/contracts/issues/887 😄

package.json
pcarranzav: Oh, looks like main is at 5.3.3 already, so I suppose this'll be 5.3.4 after rebasing

test/governance/governed.test.ts
pcarranzav: iirc the automine off default was made to support the local network setup (https://github.com/edgeandnode/local-network) but afaict it's not needed anymore (I see that repo actually sets automine to true).

So I think we should keep the previous behavior and turning this on in the fixture load function or even better in hardhat.config.ts, as the tests are designed to run with automine on (I suspect something is still turning on automine now? otherwise CI would take a lot longer to run?) and the 13s block time could cause race conditions. We used to see tests randomly failing because of this.

tmigone: Ok I've changed the `hardhat` network so that it automines by default. Everything seems to be working 💪

Maikol (dismissed):
Nice work 🚀

pcarranzav (dismissed):
LGTM (sorry for the merge conflicts, I can re-approve when you fix them)

hardhat.config.ts
```diff
@@ -3,7 +3,7 @@ import fs from 'fs'
 import * as dotenv from 'dotenv'
 import { execSync } from 'child_process'
 
-import 'hardhat/types/runtime'
+// import 'hardhat/types/runtime'
```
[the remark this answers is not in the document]
tmigone: Honestly I don't know 😵‍💫 
It's obviously also not required so I will delete it.

tmigone:
Tests fail here because fixing that would require publishing a new sdk version which is a hassle without monorepo. 
This is fixed in monorepo branch, i think it's safe to assume we merge both branches together so it shouldn't be a problem.

Maikol (dismissed):

PR #873, opened 2023-10-31
fix(GNS): use the correct token amount when curation or owner tax are zero

pcarranzav:
Before this fix, if curation or owner tax were zero, GNS would try to mint signal for the new version with zero tokens, which would revert.


tmigone (approved):
Should we also test it doesn't revert with `_curationTaxPercentage = 0`?

pcarranzav:
> Should we also test it doesn't revert with `_curationTaxPercentage = 0`?

Good point, created #891 to address in a separate PR

PR #874, opened 2023-11-02
fix: set and hardcode minimum L2 curation to 1 GRT, enforce it in all cases, allow sub-epoch collect (OZ H-01)

pcarranzav:


Maikol (approved):
Looks good to me. Do we want to update the `graph.network.yml` configs with the updated `minimumCurationDeposit` values?

contracts/l2/curation/L2Curation.sol
```diff
@@ -296,17 +284,22 @@ contract L2Curation is CurationV2Storage, GraphUpgradeable, IL2Curation {
...
-        require(
-            getCuratorSignal(curator, _subgraphDeploymentID) >= _signalIn,
-            "Cannot burn more signal than you own"
-        );
+        require(previousCuratorSignal >= _signalIn, "Cannot burn more signal than you own");
 
         // Get the amount of tokens to refund based on returned signal
         uint256 tokensOut = signalToTokens(_subgraphDeploymentID, _signalIn);
-
+        if (previousCuratorSignal > _signalIn) {
```
tmigone, line 294: pedantic nit, this could be `previousCuratorSignal != _signalIn` and save a little gas. Though not sure how you feel about code readability.

tmigone (approved):

contracts/l2/curation/L2Curation.sol
pcarranzav: Good point, but yeah for now I'll leave it like this for readability

PR #875, opened 2023-11-06
feat: sepolia staging testnet

Maikol:
Created `addresses-staging.json` for staging testnet. 
Split `etherscan.apiKey` so we can use different env variables for different networks.

### DIFF
--- a/hardhat.config.ts
+++ b/hardhat.config.ts
@@ -189,10 +189,19 @@ const config: HardhatUserConfig = {
     fork: process.env.FORK === 'true',
   },
   etherscan: {
-    apiKey: process.env.ETHERSCAN_API_KEY,
+    apiKey: {
+      mainnet: process.env.ETHERSCAN_API_KEY,
+      rinkeby: process.env.ETHERSCAN_API_KEY,
+      goerli: process.env.ETHERSCAN_API_KEY,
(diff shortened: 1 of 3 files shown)


hardhat.config.ts
```diff
@@ -189,10 +189,19 @@ const config: HardhatUserConfig = {
...
+      kovan: process.env.ETHERSCAN_API_KEY,
+      sepolia: process.env.ETHERSCAN_API_KEY,
+      arbitrumOne: process.env.ARBISCAN_API_KEY,
+      arbitrumGoerli: process.env.ARBISCAN_API_KEY,
+      arbitrumSepolia: process.env.ARBISCAN_API_KEY,
+    },
     customChains: [
       {
-        network: 'arbitrum-sepolia',
+        network: 'arbitrumSepolia',
```
tmigone, line 204: I guess this `network` refers to `etherscan.apiKey` field right? 
Just want to make sure because the convention for hardhat networks we are using would be `arbitrum-sepolia`

tmigone (approved):

hardhat.config.ts
Maikol: Yes, hardhat still uses `arbitrum-sepolia` the problem was that for etherscan apiKey we couldn't use the `arbitrum-sepolia`.

PR #888, opened 2023-11-28
chore: convert repo into a yarn workspaces monorepo and add sdk package

tmigone:
⚠️ You'll need Yarn v4 for this!! Follow readme instructions.

### Changes
- Converted repo into a yarn workspaces monorepo
- Moved contracts files into contracts package
- Updated github workflows
- Added sdk package from commit [`7d238087f93305b263f2105487127139cf2ac71c`](https://github.com/graphprotocol/sdk/tree/7d238087f93305b263f2105487127139cf2ac71c)


README.md
```diff
@@ -1,132 +1,70 @@
...
-- If MNEMONIC is set you will have available the set of addresses derived from the seed.
-- If PRIVATE_KEY is set, just that account is imported. MNEMONIC always takes precedence over PRIVATE_KEY.
-- If no MNEMONIC or PRIVATE_KEY is set it will use the remote accounts from the provider node.
-- You can always get an account using `ethers.getSigner(<address>)`
-
-Considerations when forking a chain:
+## Documentation
 
-- When running on the `localhost` network it will use by default a deterministic seed for testing purposes. If you want to connect to a local node that is forking while retaining the capability to impersonate accounts or use local accounts you need to set the FORK=true environment variable.
+> Coming soon
```
pcarranzav, line 65: Maybe we could add something like "look at each package's docs in the meantime"?

packages/contracts/contracts/curation/Curation.sol
```diff
@@ -113,11 +113,9 @@ contract Curation is CurationV2Storage, GraphUpgradeable {
      * @notice Update the minimum deposit amount to `_minimumCurationDeposit`
      * @param _minimumCurationDeposit Minimum amount of tokens required deposit
      */
-    function setMinimumCurationDeposit(uint256 _minimumCurationDeposit)
-        external
-        override
-        onlyGovernor
-    {
+    function setMinimumCurationDeposit(
```
pcarranzav, line 116: hmm is it possible that the linter config broke with the move? seems it's changing everything to a different format, but the .prettierrc.json hasn't changed, so maybe it's not picking it up anymore?

packages/contracts/package.json
```diff
@@ -0,0 +1,142 @@
+{
+  "name": "@graphprotocol/contracts",
+  "version": "5.3.2",
```
pcarranzav, line 3: After you fix conflicts in #872, worth double checking we don't actually set an older version here

packages/sdk/addresses.json
pcarranzav, line 1: Do we have duplicate addresses.json on purpose? Could the sdk import the addresses.json from the contracts package?

README.md
tmigone: good idea!

packages/sdk/addresses.json
tmigone: the address book in the sdk repo is only for running an old test script in that repo, totally not needed. deleting it.

packages/contracts/contracts/curation/Curation.sol
tmigone: mmm looks like you are right. fixing.

tmigone: Ok found the root cause. When converting to a monorepo I deleted yarn lock file to bump dependencies, it looks like newer versions of our linting tools just have different rules. 

If you delete `yarn.lock` on the `main` branch and do a fresh install then lint the code it will output the code linted as in this PR. 

So I'm inclined to keep this as is.

Maikol (approved):

pcarranzav (approved):

PR #893, opened 2023-12-20
test(e2e): use offchainlabs nitro nodes repository

tmigone:


tmigone:
See https://github.com/graphprotocol/contracts/pull/895

PR #911, opened 2024-01-11
fix: override type in CLI to allow wrapping contract calls

pcarranzav:


Maikol (dismissed):

tmigone (changes requested):
Nice! 

Can you also make this change in the SDK? `packages/sdk/src/deployments/lib/contracts/tx-log.ts`

tmigone:
Btw this code duplication problem should be resolved once we get rid of the cli folder.

packages/sdk/src/deployments/lib/contracts/tx-log.ts
```diff
@@ -66,7 +71,7 @@ export function wrapCalls(contract: Contract, contractName: string): Contract {
     wrappedContract.functions[fn] = override
```
tmigone, line 71: We should remove:
```
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
```

tmigone (approved):
Approving so you can merge but left one comment!

tmigone (approved):

PR #914, opened 2024-01-12
chore: remove cli from protocol contracts

tmigone:
### Changes
- Remove CLI from the project
- Add contract deploy hardhat task

There are a few other cli commands that don't exist as hardhat tasks but I don't think we use them. I'd rather only migrate them if/when we need them.


Maikol (dismissed):
🎉

Maikol (approved):

PR #939, opened 2024-02-02
fix: clear vote for earlier rounds (OZ L-02)

Maikol:


packages/contracts/contracts/rewards/SubgraphAvailabilityManager.sol
```diff
@@ -203,6 +203,12 @@ contract SubgraphAvailabilityManager is Governed {
             : lastAllowVote[currentNonce][_subgraphDeploymentID];
         lastVoteForSubgraph[_oracleIndex] = timestamp;
 
+        // clear opposite vote for a subgraph deployment if it exists
+        uint256[NUM_ORACLES] storage oppositeVoteForSubgraph = _deny
```
tmigone, line 207: I think we could combine the two with an if/else statement to avoid testing twice for the value of `_deny`. And that would also eliminate the need for the storage variables `lastVote...` and `oppositeVote...` (i think).

tmigone (changes requested):

tmigone (approved):

PR #944, opened 2024-02-20
❄️  Graph Horizon and Subgraph Service ❄️

pcarranzav:


packages/contracts/contracts/arbitrum/ITokenGateway.sol
```diff
@@ -47,20 +47,20 @@ interface ITokenGateway {
     // );
 
     function outboundTransfer(
-        address _token,
-        address _to,
-        uint256 _amount,
-        uint256 _maxGas,
-        uint256 _gasPriceBid,
-        bytes calldata _data
+        address oken,
```
pcarranzav, line 50: nit: not sure why these got changed, but this should be `token` and the one below should be `to`

packages/contracts/contracts/rewards/RewardsManager.sol
```diff
@@ -349,29 +360,34 @@ contract RewardsManager is RewardsManagerV4Storage, GraphUpgradeable, IRewardsMa
...
+        (
+            address indexer,
+            bytes32 subgraphDeploymentID,
+            uint256 tokens,
+            uint256 accRewardsPerAllocatedToken
+        ) = IRewardsIssuer(rewardsIssuer).getAllocationData(_allocationID);
 
-        IStaking.Allocation memory alloc = staking.getAllocation(_allocationID);
-        uint256 accRewardsPerAllocatedToken = onSubgraphAllocationUpdate(alloc.subgraphDeploymentID);
+        uint256 updatedAccRewardsPerAllocatedToken = onSubgraphAllocationUpdate(subgraphDeploymentID);
```
pcarranzav, line 373: This will call `getAccRewardsPerAllocatedToken` which is still getting the amount of allocated tokens from staking only. I think `getAccRewardsPerAllocatedToken` should be updated to also fetch from the Subgraph Service and add them?

packages/horizon/README.md
```diff
@@ -0,0 +1,13 @@
+# Sample Hardhat Project
+
+This project demonstrates a basic Hardhat use case. It comes with a sample contract, a test for that contract, and a script that deploys that contract.
```
pcarranzav, line 3: We're far from a basic hardhat use case now 🤣 
(let's make a note to update this README in a future PR)

pcarranzav (commented):
A few comments for now - still reviewing

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
```diff
@@ -0,0 +1,142 @@
...
+ * using a Horizon provision. See {IDataServiceFees} for more details.
+ */
+abstract contract DataServiceFees is DataService, DataServiceFeesV1Storage, IDataServiceFees {
+    using ProvisionTracker for mapping(address => uint256);
+    using LinkedList for LinkedList.List;
+
+    /**
+     * @notice See {IDataServiceFees-releaseStake}
+     */
+    function releaseStake(uint256 n) external virtual {
```
pcarranzav, line 25: shouldn't this have `override`?

packages/horizon/contracts/libraries/LinkedList.sol
```diff
@@ -0,0 +1,143 @@
...
+        bytes32 nextItem = getNextItem(self.head);
+        deleteItem(self.head);
+        self.count -= 1;
+        self.head = nextItem;
+        if (self.count == 0) self.tail = bytes32(0);
+        return self.head;
+    }
+
+    /**
+     * @notice Traverses the list and processes each item.
```
pcarranzav, line 98: Worth mentioning that it processes and also potentially deletes each processed item

packages/horizon/contracts/libraries/LinkedList.sol
```diff
@@ -0,0 +1,143 @@
...
+     * @param self The list metadata
+     * @param getNextItem A function to get the next item in the list. It should take
+     * the id of the current item and return the id of the next item.
+     * @param processItem A function to process an item. The function should take the id of the item
+     * and an accumulator, and return:
+     * - a boolean indicating whether the traversal should stop
+     * - a boolean indicating whether the item should be deleted
```
pcarranzav, line 105: This doesn't seem to be implemented, right? Deletion is not optional?

packages/horizon/contracts/libraries/LinkedList.sol
```diff
@@ -0,0 +1,143 @@
...
+     * - an accumulator to pass data between iterations
+     * @param deleteItem A function to delete an item. This should delete the item from
+     * the contract storage. It takes the id of the item to delete.
+     * @param processInitAcc The initial accumulator data
+     * @param iterations The maximum number of iterations to perform. If 0, the traversal will continue
+     * until the end of the list.
+     */
+    function traverse(
```
pcarranzav, line 113: This is beautiful, nicely done 👏

pcarranzav (commented):
A few more comments, still reviewing...

packages/horizon/contracts/data-service/interfaces/IDataService.sol
```diff
@@ -0,0 +1,139 @@
...
+     * @notice Emitted when a service provider accepts a provision in {Graph Horizon staking contract}.
+     * @param serviceProvider The address of the service provider.
+     */
+    event ProvisionAccepted(address indexed serviceProvider);
+
+    /**
+     * @notice Emitted when a service provider starts providing the service.
+     * @param serviceProvider The address of the service provider.
+     */
+    event ServiceStarted(address indexed serviceProvider);
```
pcarranzav, line 33: Same here, providing additional data would allow indexing more information about the service

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * @notice Checks if a provision of a service provider is valid according
+     * to the parameter ranges established.
+     */
+    modifier onlyValidProvision(address serviceProvider) virtual {
+        IHorizonStaking.Provision memory provision = _getProvision(serviceProvider);
+        _checkProvisionTokens(provision);
+        _checkProvisionDelegationRatio(provision, serviceProvider);
```
pcarranzav, line 94: Since we're not checking this in SubgraphService, I kinda think we should remove the delegation ratio check from here as well... I'd expect other data services to use delegation ratio in the same way as SubgraphService, using the ratio when computing available stake

packages/horizon/contracts/interfaces/IBridgeEscrow.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+pragma solidity 0.8.26;
+
+/**
+ * @title IBridgeEscrow
+ */
+interface IBridgeEscrow {
```
pcarranzav, line 8: BridgeEscrow only exists on the L1 protocol, so I think we could omit it here and in GraphDirectory? Since the L2 contracts won't have an entry for it, and we won't deploy Horizon on L1.

packages/horizon/contracts/libraries/Denominations.sol
```diff
@@ -0,0 +1,21 @@
...
+pragma solidity 0.8.26;
+
+/**
+ * @title Denominations library
+ * @dev Provides a list of ground denominations for those tokens that cannot be represented by an ERC20.
+ * For now, the only needed is the native token that could be ETH, MATIC, or other depending on the layer being operated.
+ */
+library Denominations {
+    /// @notice The address of the native token, i.e ETH
+    address internal constant NATIVE_TOKEN = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
```
pcarranzav, line 11: nit: Worth noting in comments that this comes from ERC-7528

pcarranzav (commented):
Moar

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -0,0 +1,673 @@
...
+     * @param _deposit Amount of tokens staked as deposit
+     * @param _allocationId Allocation disputed
+     */
+    function _createIndexingDisputeWithAllocation(
+        address _fisherman,
+        uint256 _deposit,
+        address _allocationId
+    ) private returns (bytes32) {
+        // Create a disputeId
+        bytes32 disputeId = keccak256(abi.encodePacked(_allocationId));
```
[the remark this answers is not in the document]
tmigone: hmm how about adding the POI?

packages/contracts/contracts/rewards/RewardsManager.sol
tmigone: fixed here: https://github.com/graphprotocol/contracts/pull/982

packages/horizon/contracts/data-service/DataService.sol
```diff
@@ -0,0 +1,51 @@
...
+    function __DataService_init() internal onlyInitializing {
+        __DataService_init_unchained();
+    }
+
+    /**
+     * @notice Initializes the contract.
+     */
+    // solhint-disable-next-line func-name-mixedcase
+    function __DataService_init_unchained() internal onlyInitializing {
+        __ProvisionManager_init_unchained();
```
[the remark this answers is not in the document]
tmigone: yes good catch, that init call should be in the other function. Its all a bit confusing tbh, but seems necessary.

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
```diff
@@ -0,0 +1,142 @@
...
+ * @dev Implementation of the {IDataServiceFees} interface.
+ * @notice Extension for the {IDataService} contract to handle payment collateralization
+ * using a Horizon provision. See {IDataServiceFees} for more details.
+ */
+abstract contract DataServiceFees is DataService, DataServiceFeesV1Storage, IDataServiceFees {
+    using ProvisionTracker for mapping(address => uint256);
+    using LinkedList for LinkedList.List;
+
+    /**
+     * @notice See {IDataServiceFees-releaseStake}
```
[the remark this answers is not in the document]
tmigone: oh man. That's useful. I was copying OZ's style a bit too much and they don't seem to use it. 
I'll make a note to change it after audit starts.

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
tmigone: yep good catch. since it's an abstract contract the lack of override wasn't showing as an error

packages/horizon/contracts/libraries/LinkedList.sol
tmigone: Right, that's old docs. 
The first implementation allowed you to traverse it in "read only" mode, so you had that parameter. But turns out is far easier to just loop the elements with a while loop if you don't care about deleting/processing them.

packages/contracts/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @graphprotocol/contracts
 
+## 7.0.0
+
+### Major Changes
+
+- 9686ce1: Add Horizon staking interface
```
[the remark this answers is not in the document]
tmigone: haha. yes we need to update this.

PR #954, opened 2024-02-23
chore: update token-distribution to use our linting packages

tmigone:


Maikol (approved):
LGTM

packages/token-distribution/test/tokenLock.test.ts
```diff
@@ -324,7 +325,8 @@ describe('GraphTokenLockSimple', () => {
               // Before cliff no vested tokens
               if (cliffTime.gt(0) && currentTime.lt(cliffTime)) {
                 expectedAmount = BigNumber.from(0)
-              } else {
+              }
+              else {
```
pcarranzav, line 329: This looks like a weird choice for linter output doesn't it? `} else {` looked cleaner imo, not sure if that's configurable

pcarranzav (dismissed):
One detail but LGTM

packages/token-distribution/test/tokenLock.test.ts
tmigone: yeah i actually agree with you, going to look into it, I didn't notice this.

Maikol (approved):

PR #963, opened 2024-03-19
chore: add DataEdge contracts

Maikol:
Moved `DataEdge` contract from [Block Oracle](https://github.com/graphprotocol/block-oracle/tree/main/packages/contracts) repo.


packages/data-edge/.gitattributes
```diff
@@ -0,0 +1 @@
+*.sol linguist-language=Solidity
```
tmigone, line 1: We can safely remove this file, it's from a time where GitHub had trouble with Solidity

packages/data-edge/.husky/commit-msg
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
tmigone, line 1: These husky files should be at the repo level not package. We already have one `commit-msg` which is identical to this one so we can delete it.

packages/data-edge/.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
tmigone, line 1: For this one you can edit `.husky/pre-commit` and add it there.

tmigone (changes requested):
Left some comments!

Would be nice to change the linting in the package to use `eslint-graph-config` and `solhint-graph-config` :eyes: I dont think is absolutely necessary but shouldn't be too hard.

tmigone (changes requested):

tmigone (approved):

PR #968, opened 2024-04-12
[WIP] Horizon: add escrow and payments

tmigone:


packages/horizon/contracts/escrow/GraphEscrow.sol
```diff
@@ -0,0 +1,221 @@
...
+        if (_withdrawEscrowThawingPeriod > MAX_THAWING_PERIOD) {
+            revert GraphEscrowThawingPeriodTooLong(_withdrawEscrowThawingPeriod, MAX_THAWING_PERIOD);
+        }
+
+        revokeCollectorThawingPeriod = _revokeCollectorThawingPeriod;
+        withdrawEscrowThawingPeriod = _withdrawEscrowThawingPeriod;
+    }
+
+    // approve a data service to collect funds
+    function approveCollector(address dataService, uint256 amount) external {
```
tmigone, line 61: A payer could bypass the thawing by re-approving a collector for a 0 amount. 

I think we should not allow re approvals for amounts that are smaller than the current allowance. 

If you want to reduce the allowance it should go through the thawing period.

PR #976, opened 2024-05-27
fix: slashing horizon delegation

Maikol:


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -387,7 +387,9 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
     ) external override notPaused {
         address verifier = msg.sender;
         Provision storage prov = _provisions[serviceProvider][verifier];
-        require(prov.tokens >= tokens, HorizonStakingInsufficientTokens(prov.tokens, tokens));
+        DelegationPoolInternal storage pool = _getDelegationPool(serviceProvider, verifier);
+        uint256 availableTokens = prov.tokens + pool.tokens;
```
tmigone, line 391: Nit, I would rename this variable just to avoid any potential confusion with actual `tokensAvailable` in a provision. Maybe `tokensSlashable` or `tokensProvisionTotal`?

Maikol: yeah, I like `tokensProvisionTotal`, will update.

PR #993, opened 2024-08-15
fix: change pending rewards due to resize to token amounts instead of per token (OZ H-01)

tmigone:


Maikol (approved):

packages/contracts/contracts/rewards/IRewardsIssuer.sol
```diff
@@ -10,13 +10,20 @@ interface IRewardsIssuer {
      * @return subgraphDeploymentId Subgraph deployment id for the allocation
      * @return tokens Amount of allocated tokens
      * @return accRewardsPerAllocatedToken Rewards snapshot
+     * @return accRewardsPending Tokens pending to be claimed
```
pcarranzav, line 13: nit: the description here isn't super clear. Maybe something like "Snapshot of accumulated rewards from previous allocation resizing, pending to be claimed"?

PR #994, opened 2024-08-15
fix: early return in Staking withdraw (OZ M-03)

Maikol:


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -544,8 +544,8 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
             emit StakeWithdrawn(serviceProvider, _tokens);
         } else {
             // Before locking more tokens, withdraw any unlocked ones if possible
-            if (sp.__DEPRECATED_tokensLockedUntil != 0 && block.number >= sp.__DEPRECATED_tokensLockedUntil) {
```
tmigone, line 547: Can we have `p.__DEPRECATED_tokensLockedUntil = 0` and `sp.__DEPRECATED_tokensLocked != 0`? I think not no?

tmigone (approved):

packages/horizon/contracts/staking/HorizonStaking.sol
Maikol: a good question, I assume that's not possible right?

PR #996, opened 2024-08-15
fix: add PaymentsEscrow functions for collector tracking (OZ M-04)

Maikol:


tmigone (approved):
I don't mind us adding these getters but we already would have derived the information from events and into a subgraph which is what offchain components would use. 

Then the service providers when doing business with a payer (gateway) should consider thawing collectors as revoked. Meaning they should not choose to do business with them.

tmigone:
Actually, `authorizedCollectors` is already public so it has a getter. I would remove these then I don't think they serve much purpose.

PR #1002, opened 2024-08-20
fix: make fisherman deposit a fixed value (OZ L-01)

Maikol:


packages/subgraph-service/contracts/interfaces/IDisputeManager.sol
```diff
@@ -151,12 +151,11 @@ interface IDisputeManager {
     error DisputeManagerDisputePeriodZero();
     error DisputeManagerZeroTokens();
     error DisputeManagerInvalidDispute(bytes32 disputeId);
-    error DisputeManagerInvalidMinimumDeposit(uint256 minimumDeposit);
+    error DisputeManagerInvalidDisputeDeposit(uint256 disputeDeposit);
```
tmigone, line 154: I think we can make this error a bit more specific
```suggestion
    error DisputeManagerInvalidZeroDeposit(uint256 disputeDeposit);
```

tmigone (changes requested):
I think we don't need the function `_pullSubmitterDeposit()`, all its doing is pulling tokens so I think it's best if we remove it and just pull using `_graphToken().pullTokens(msg.sender, disputeDeposit);`

PR #1015, opened 2024-08-23
fix: add fn to allow force closing stale allocations (OZ H-09) 

tmigone:


packages/subgraph-service/contracts/libraries/Allocation.sol
```diff
@@ -164,6 +166,16 @@ library Allocation {
         return _get(self, allocationId);
     }
 
+    /**
+     * @notice Checks if an allocation is stale
+     * @param self The allocation
+     * @param staleThreshold The time in blocks to consider an allocation stale
+     */
+    function isStale(State memory self, uint256 staleThreshold) internal view returns (bool) {
+        uint256 timeSinceLastPOI = block.number - Math.max(self.createdAt, self.lastPOIPresentedAt);
```
pcarranzav, line 175: createdAt is a timestamp, so this should be block.timestamp instead of block.number right?

PR #1025, opened 2024-09-10
feat(horizon): deploy horizon with Hardhat Ignition

tmigone:


packages/horizon/ignition/modules/core/HorizonProxies.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  // Deploy HorizonStaking proxy without an implementation
+  const HorizonStakingProxy = m.contract('GraphProxy', GraphProxyArtifact, [ZERO_ADDRESS, GraphProxyAdmin], { after: [PeripheryRegistered], id: 'GraphProxy_HorizonStaking' })
+
+  // Deploy proxies for payments contracts using OZ TransparentUpgradeableProxy
+  const { Proxy: GraphPaymentsProxy, ProxyAdmin: GraphPaymentsProxyAdmin } = deployWithOZProxy(m, 'GraphPayments')
+  const { Proxy: PaymentsEscrowProxy, ProxyAdmin: PaymentsEscrowProxyAdmin } = deployWithOZProxy(m, 'PaymentsEscrow')
+
+  // Register the proxies in the controller
+  const setProxyHorizonStaking = m.call(Controller, 'setContractProxy', [ethers.keccak256(ethers.toUtf8Bytes('Staking')), HorizonStakingProxy], { id: 'setContractProxy_HorizonStaking' })
```
alcuadrado, line 25: I believe this hashing is a custom pattern, is it?

packages/horizon/ignition/modules/core/HorizonStaking.ts
```diff
@@ -0,0 +1,34 @@
...
+      after: [PeripheryRegistered, HorizonRegistered],
+    },
+  )
+
+  // Upgrade proxy to implementation contract
+  const upgradeCall = m.call(GraphProxyAdmin, 'upgrade', [HorizonStakingProxy, HorizonStakingImplementation])
+  const acceptCall = m.call(GraphProxyAdmin, 'acceptProxy', [HorizonStakingImplementation, HorizonStakingProxy], { after: [upgradeCall] })
+
+  // Load contract with implementation ABI
+  const HorizonStaking = m.contractAt('HorizonStaking', HorizonStakingProxy, { after: [acceptCall], id: 'HorizonStaking_Instance' })
```
alcuadrado, line 31: Is `contractAt` forcing you to use explicit `id`s too often? Or are you setting it here for a stylistic preference?

packages/horizon/ignition/modules/periphery/EpochManager.ts
```diff
@@ -0,0 +1,19 @@
+import { buildModule } from '@nomicfoundation/hardhat-ignition/modules'
+import { deployWithGraphProxy } from '../proxy/GraphProxy'
+
+import ControllerModule from './Controller'
+import EpochManagerArtifact from '@graphprotocol/contracts/build/contracts/contracts/epochs/EpochManager.sol/EpochManager.json'
```
alcuadrado, line 5: I noticed these imports in a few files. Are these npm packages that you had already published? Are these contracts verified already? I'm mostly asking to understand:

1. Are you loading from `node_module` or from this monorepo? If the latter, why?
2. Should we do anything with its build info/sources so that then it can be verified easily?

packages/horizon/ignition/modules/periphery/GraphToken.ts
```diff
@@ -0,0 +1,33 @@
...
+  const governor = m.getParameter('governor')
+  const initialSupply = m.getParameter('initialSupply')
+
+  const { instance: GraphToken } = deployWithGraphProxy(m, {
+    name: 'GraphToken',
+    artifact: GraphTokenArtifact,
+    args: [deployer],
+  })
+
+  // TODO: move this mint to a testnet only module
```
alcuadrado, line 24: How have you been managing testnet/mainnet deployment differences so far?

alcuadrado (commented):
This is looking really good, @tmigone! While I didn't spend enough time to understand the architecture of Horizon, I looked for unsage patterns and the structure of the modules, and it does look like what we expect to see in a project this size.

If left a few comments, including a few questions about things we'd love to know to see if/how we can make Ignition better.

packages/horizon/ignition/modules/core/HorizonStaking.ts
alcuadrado: Oh, I guess this happens when you do "deploy implementation", "contractAt implementation". I'll think about how to improve this. 

Maybe with something like what you did here: https://github.com/graphprotocol/contracts/pull/1025#discussion_r1761950531

packages/horizon/ignition/configs/graph.hardhat.json
```diff
@@ -0,0 +1,45 @@
+{
+  "GraphProxyAdmin": {
+    "governor": "0xFFcf8FDEE72ac11b5c542428B35EEF5769C409f0"
```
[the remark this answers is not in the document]
tmigone: Yes exactly!

packages/horizon/ignition/modules/core/GraphPayments.ts
```diff
@@ -0,0 +1,28 @@
...
+  const { Controller, PeripheryRegistered } = m.useModule(GraphPeripheryModule)
+  const { GraphPaymentsProxyAdmin, GraphPaymentsProxy, HorizonRegistered } = m.useModule(HorizonProxiesModule)
+
+  const protocolPaymentCut = m.getParameter('protocolPaymentCut')
+
+  // Deploy GraphPayments implementation
+  const GraphPaymentsImplementation = m.contract('GraphPayments',
+    [Controller, protocolPaymentCut],
+    {
+      after: [PeripheryRegistered, HorizonRegistered],
```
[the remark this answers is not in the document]
tmigone: That would be sweet 🙏

packages/horizon/scripts/deploy.ts
```diff
@@ -0,0 +1,16 @@
+import { ignition } from 'hardhat'
```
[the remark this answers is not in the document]
tmigone: I think I initially added this because one of the contract calls during setup was reverting and I wanted to log some stuff, so I should probably remove it.

But now that I think of it, there is one potential use case where it might be handy. For our products and internal tooling we use a contract address book with a specific format which is different from the one ignition outputs. Changing everything to support ignition's format seems more complicated than just adding a post-deployment processing step to convert between formats. Though ... perhaps it's best to keep that post processing on an isolated hardhat task and run ignition through CLI.

packages/horizon/ignition/modules/proxy/TransparentUpgradeableProxy.ts
```diff
@@ -0,0 +1,32 @@
+import { IgnitionModuleBuilder } from '@nomicfoundation/ignition-core'
+
+import ProxyAdminArtifact from '@openzeppelin/contracts/build/contracts/ProxyAdmin.json'
+import TransparentUpgradeableProxyArtifact from '@openzeppelin/contracts/build/contracts/TransparentUpgradeableProxy.json'
+
+// Deploy a TransparentUpgradeableProxy
+// Note that this module uses a dummy contract as the implementation as the proxy requires a valid contract
+// address to be deployed
+export function deployWithOZProxy(m: IgnitionModuleBuilder, contractName: string) {
+  const deployer = m.getAccount(0)
```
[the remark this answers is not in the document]
tmigone: I think it was fairly straightforward to realize. Definitely nitpicking but maybe you could mention this pattern in this section of the docs: https://hardhat.org/ignition/docs/guides/creating-modules#future--ids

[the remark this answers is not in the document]
tmigone: well now that i re-read it, the existing docs already sort of cover it, so not sure if it's worth further clarifying. Perhaps would be better as an example in the examples repo?

packages/horizon/ignition/modules/proxy/TransparentUpgradeableProxy.ts
```diff
@@ -0,0 +1,32 @@
...
+
+  const dummy = m.contract('Dummy', [], { id: `OZProxyDummy_${contractName}` })
```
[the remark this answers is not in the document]
tmigone: I don't think there is anything Ignition could do better here, it's really a limitation of the OZ proxy not accepting 0x00 as a valid implementation address (or an address without code deployed to it).

Agreed that using the implementation future is cleaner, and in some other instances that's what we are doing (for example for GraphProxy), but in this specific case there is sort of a circular dependency here where the implementation itself needs to know the proxy address before it's deployed, but at the same time the proxy needs the implementation address... Again this is a combination of OZ proxies and some Horizon weirdness (which I hope we can fix in the future).

packages/horizon/ignition/modules/core/HorizonProxies.ts
tmigone: yes!

packages/horizon/ignition/modules/core/HorizonStaking.ts
tmigone: yeah had to set it because of that! 

Perhaps Ignition could detect this pattern:
```
const proxy = m.contract('Proxy')
const implementation = m.contract('ContractName')
const proxyWithImplementationABI = m.contractAt('ContractName', proxy)
```

And automatically add something to avoid id collision on the contractAt. But it's also not a big deal I think.

packages/horizon/ignition/modules/periphery/EpochManager.ts
tmigone: Good question. So, Horizon project (`packages/horizon`) relies on some contracts from the current version of the protocol (`packages/contracts`) which remain mostly unchanged. In order to be able to use modern tooling and make it easier for auditors we decided to spin up a new package (`horizon`) where we put all the new stuff then eventually the old stuff that's required will be moved to `horizon` and `contracts` will be deprecated.

So to answer your questions, anything we are pulling from `@graphprotocol/contracts` is being pulled from the monorepo package `contracts` and those are already deployed to production and verified. I don't know what would happen if I tried to verify something from the `contracts` package from within `horizon` 🤔 but in our specific case that will never happen as anything we pull from there should already be verified.

packages/horizon/ignition/modules/periphery/GraphToken.ts
tmigone: I was actually wondering that. So far I have not been managing it at all. Everything so far assumes it's a testnet deployment, I'm curious what patterns you've seen for this? 

I can think of two approaches:
1. Add testnet/mainnet modules with specific config/steps for each environment. Those modules would import the main Horizon module, deploy everything in there then add on top of that.
2. Deploy with a script instead of CLI and have any logic in there to deal with differences.

packages/horizon/scripts/deploy.ts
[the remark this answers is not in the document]
alcuadrado: Maybe we could have a subtask that's run after a successful Ignition deployment, where you can run your data transformation. WDYT?

/cc @zoeyTM

packages/horizon/ignition/modules/proxy/TransparentUpgradeableProxy.ts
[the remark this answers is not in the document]
alcuadrado: I created an issue, as I think we could have a short guide about this: https://github.com/NomicFoundation/hardhat-ignition/issues/813

packages/horizon/ignition/modules/proxy/TransparentUpgradeableProxy.ts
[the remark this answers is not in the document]
alcuadrado: I see. Happy to help in any way if you want to explore an alternative in the future.

packages/horizon/ignition/modules/core/HorizonStaking.ts
alcuadrado: Thanks for confirming, I created an issue to keep track of this: https://github.com/NomicFoundation/hardhat-ignition/issues/814

packages/horizon/ignition/modules/periphery/EpochManager.ts
alcuadrado: Ok, I think we are good for now. But it's an interesting use case. I think we may want to let you provide their build infos. Do you keep track of them?

packages/horizon/ignition/modules/periphery/GraphToken.ts
alcuadrado: Another idea that we had explored was allowing the users to create "chain specific" modules, that would behave different depending on the chainId. Imagine something like this:

```
const mod = chainSpecificModule("Mod", 
{
   1: (m) => {
      // mainnet logic
   },
   31337: (m) => {
      // hardhat network logic
   }
})
```

(ignore the horrible syntax)

This could work, especially if all you do are operations inside of those modules. If you need to export contracts, maybe we'd have to force you to export the same contract futures or something like that, so that it plays nicely with typescript.

Do you normally do ops there?

alcuadrado: Another pattern would be something like having your `main` module, and then for testnets you have `testnet` modules, which `useModule(main)`, and run those ops. This feels closer to extending a class for testing purposes.

The disadvantage there is that is may force you to structure `main` in a certain way so that you can do those ops.

packages/horizon/scripts/deploy.ts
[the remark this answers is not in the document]
tmigone: yeah that sounds great

packages/horizon/ignition/modules/periphery/EpochManager.ts
tmigone: Not sure I follow! Keep track of what?

packages/horizon/ignition/modules/periphery/GraphToken.ts
tmigone: That last pattern you described is what I had in mind. Chain specific modules would be really helpful though, there are certain operations for which I think it makes a lot of sense (for example, minting initial supply of tokens on a testnet)

PR #1027, opened 2024-09-16
chore: add stakeToFeesRatio to SubgraphService initialize

Maikol:


packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -372,9 +377,8 @@ contract SubgraphService is
     /**
      * @notice See {ISubgraphService.setStakeToFeesRatio}
      */
-    function setStakeToFeesRatio(uint256 stakeToFeesRatio_) external override onlyOwner {
-        stakeToFeesRatio = stakeToFeesRatio_;
-        emit StakeToFeesRatioSet(stakeToFeesRatio_);
+    function setStakeToFeesRatio(uint256 stakeToFeesRatio) external override onlyOwner {
```
tmigone, line 380: I'd keep the `_` here just in case

tmigone (changes requested):

packages/subgraph-service/contracts/SubgraphService.sol
Maikol: yeah, I was on the fence but makes sense, updated!

tmigone (approved):
💠

PR #1029, opened 2024-09-17
fix: remove unused errors, use LegacyAllocationAlreadyMigrated to prevent migrating a legacy allocation twice (OZ N-02)

MoonBoi9001:
### Motivation:

- This PR addresses the [following N-02 OZ audit issue](https://defender.openzeppelin.com/#/audit/9d56f9fe-e25e-4ac5-acb0-772150889078/issues/N-02), applying the recommended fixes.

### Title: 
N-02 Unused Errors

### Details: 
Unused errors can negatively affect code intent and readability.

The [AllocationManagerZeroTokensAllocation](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/subgraph-service/contracts/utilities/AllocationManager.sol#L140), [DataServiceFeatureNotImplemented](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/data-service/interfaces/IDataService.sol#L66), and [LegacyAllocationAlreadyMigrated](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/subgraph-service/contracts/libraries/LegacyAllocation.sol#L37) errors are unused.

Consider deleting the AllocationManagerZeroTokensAllocation error since creating allocations with no tokens is allowed for altruistic allocations. In addition, consider deleting the DataServiceFeatureNotImplemented error and using LegacyAllocationAlreadyMigrated to prevent migrating a legacy allocation twice.

##

### Key changes

- removed unused AllocationManagerZeroTokensAllocation 
- removed unused DataServiceFeatureNotImplemented
- now using LegacyAllocationAlreadyMigrated to prevent migrating a legacy allocation twice

### DIFF
--- a/packages/subgraph-service/contracts/libraries/LegacyAllocation.sol
+++ b/packages/subgraph-service/contracts/libraries/LegacyAllocation.sol
@@ -39,11 +39,12 @@ library LegacyAllocation {
     /**
      * @notice Migrate a legacy allocation
      * @dev Requirements:
-     * - The allocation must not exist
+     * - The allocation must not have been previously migrated
      * @param self The legacy allocation list mapping
      * @param indexer The indexer that owns the allocation
      * @param allocationId The allocation id
      * @param subgraphDeploymentId The subgraph deployment id the allocation is for
+     * @custom:error LegacyAllocationAlreadyMigrated if the allocation has already been migrated
      */
     function migrate(
         mapping(address => State) storage self,
@@ -51,7 +52,9 @@ library LegacyAllocation {
         address allocationId,
         bytes32 subgraphDeploymentId
     ) internal {
-        require(!self[allocationId].exists(), LegacyAllocationExists(allocationId));
+        if (self[allocationId].exists()) {
+            revert LegacyAllocationAlreadyMigrated(allocationId);
+        }
 
         State memory allocation = State({ indexer: indexer, subgraphDeploymentId: subgraphDeploymentId });
         self[allocationId] = allocation;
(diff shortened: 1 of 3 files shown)


packages/subgraph-service/contracts/libraries/LegacyAllocation.sol
```diff
@@ -39,19 +39,22 @@ library LegacyAllocation {
...
     function migrate(
         mapping(address => State) storage self,
         address indexer,
         address allocationId,
         bytes32 subgraphDeploymentId
     ) internal {
-        require(!self[allocationId].exists(), LegacyAllocationExists(allocationId));
+        if (self[allocationId].exists()) {
+            revert LegacyAllocationAlreadyMigrated(allocationId);
+        }
```
tmigone, line 57: This should be equivalent and matches the error throwing pattern we are using (use require and throw custom errors).

```suggestion
        require(!self[allocationId].exists(), LegacyAllocationAlreadyMigrated(allocationId);
```

tmigone (changes requested):

packages/subgraph-service/contracts/libraries/LegacyAllocation.sol
MoonBoi9001: Okay, apparently require is more gas efficient than if revert? But happy to change this back

tmigone: To be honest I'm not sure which one is more gas efficient 😄 
Since we are on Arbitrum One where gas is no longer a huge problem, I'd rather have it be more readable regardless of the gas cost in this case.

tmigone (approved):
👍

PR #1030, opened 2024-09-17
chore(SubgraphService): add Multicall to SubgraphService

Maikol:


tmigone (commented):
We also need to call `__Multicall_init()` in the `initialize()` function (it doesn't make any difference in this case because `__Multicall_init()` is empty, but for muscle memory and best practices I think we should still do it).

tmigone (changes requested):
See comment above.

tmigone (approved):

PR #1034, opened 2024-09-18
fix: address inconsistent parameter and mapping order (OZ N-05)

MoonBoi9001:
### Motivation:

- This PR addresses the [following N-05 OZ audit issue](https://defender.openzeppelin.com/#/audit/9d56f9fe-e25e-4ac5-acb0-772150889078/issues/N-05), applying the recommended fixes.

### Title: 
N-05 Inconsistent Parameter and Mapping Order

### Details: 
In the HorizonStaking contract, the [_isAuthorized function](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStaking.sol#L949) has the following parameter order: _operator, _serviceProvider, _verifier. On the other hand, the [_operatorAuth mapping](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStakingStorage.sol#L162) uses this order: _serviceProvider, _verifier, _operator. In addition, the [_legacyOperatorAuth variable](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStakingStorage.sol#L108) is defined as a mapping from an operator to a service provider to a boolean status. However, [it is being used](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStaking.sol#L937) as a mapping from a service provider to an operator to a boolean status.

These inconsistencies can lead to confusion and potential errors during integration or modification, as developers may misinterpret the correct parameter order.

Consider aligning the parameter order in both the _isAuthorized function and the _operatorAuth mapping for improved consistency. Moreover, consider updating the keys and value names in the _legacyOperatorAuth mapping definition to match its usage.

##

### Key changes:

- Aligned the parameter order in the _isAuthorized function
- Aligned the parameter order in the isAuthorized function
- Aligned the parameter order in the _operatorAuth function
- Aligned the parameter order in the ProvisionManagerNotAuthorized function
- Aligned the parameter order in the HorizonStakingNotAuthorized function

##

### Consistent peramater order:

- ### ServiceProvider, Verifier, Operator

### DIFF
--- a/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
+++ b/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
@@ -79,7 +79,7 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
      * @param caller The address of the caller.
      * @param serviceProvider The address of the service provider.
      */
-    error ProvisionManagerNotAuthorized(address caller, address serviceProvider);
+    error ProvisionManagerNotAuthorized(address serviceProvider, address caller);
 
     /**
      * @notice Thrown when a provision is not found.
@@ -93,7 +93,7 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
     modifier onlyAuthorizedForProvision(address serviceProvider) {
         require(
             _graphStaking().isAuthorized(msg.sender, serviceProvider, address(this)),
-            ProvisionManagerNotAuthorized(msg.sender, serviceProvider)
+            ProvisionManagerNotAuthorized(serviceProvider, msg.sender)
         );
         _;
     }
--- a/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
+++ b/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
@@ -76,8 +76,8 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
 
     /**
      * @notice Thrown when the caller is not authorized to manage the provision of a service provider.
+     * @param serviceProvider The address of the serviceProvider.
      * @param caller The address of the caller.
-     * @param serviceProvider The address of the service provider.
      */
     error ProvisionManagerNotAuthorized(address serviceProvider, address caller);
 
--- a/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
+++ b/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
@@ -92,7 +92,7 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
      */
     modifier onlyAuthorizedForProvision(address serviceProvider) {
         require(
-            _graphStaking().isAuthorized(msg.sender, serviceProvider, address(this)),
+            _graphStaking().isAuthorized(serviceProvider, address(this), msg.sender),
             ProvisionManagerNotAuthorized(serviceProvider, msg.sender)
         );
         _;
--- a/packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
+++ b/packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
@@ -108,14 +108,14 @@ interface IHorizonStakingMain {
     /**
      * @dev Emitted when an operator is allowed or denied by a service provider for a particular verifier
      * @param serviceProvider The address of the service provider
-     * @param operator The address of the operator
      * @param verifier The address of the verifier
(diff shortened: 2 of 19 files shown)


MoonBoi9001:
##

### Did not address:

- > In addition, the [_legacyOperatorAuth variable](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStakingStorage.sol#L108) is defined as a mapping from an operator to a service provider to a boolean status. However, [it is being used](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStaking.sol#L937) as a mapping from a service provider to an operator to a boolean status.
  - This point was already addressed. 

 - > Moreover, consider updating the keys and value names in the _legacyOperatorAuth mapping definition to match its usage.
   - I'm unsure what changes need to happen given:
   - `mapping(address serviceProvider => mapping(address legacyOperator => bool authorized)) internal _legacyOperatorAuth;`
   - I presume the suggestion is to change `authorized` to `allowed`? [See here](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/staking/HorizonStaking.sol#L937)

tmigone (changes requested):
We should also update `isAuthorized()` external function, and probably in the `tests/` folder there are cases where we call this function and we need to switch the param order as well. For example: https://github.com/graphprotocol/contracts/blob/95b3f83b1abe85449a63ab4922e84896f989ac57/packages/horizon/test/shared/horizon-staking/HorizonStakingShared.t.sol#L758

MoonBoi9001:
Presumably this would also imply updating _getStorage_OperatorAuth

https://github.com/graphprotocol/contracts/blob/95b3f83b1abe85449a63ab4922e84896f989ac57/packages/horizon/test/shared/horizon-staking/HorizonStakingShared.t.sol#L757

tmigone:
Yes good idea being consistent. I wouldn't mind keeping `_getStorage_OperatorAuth` as is because its an internal test only function, but anything that's externally accessible it does seem like a good idea to be consistent.

MoonBoi9001:
New Proposed consistent order after discussions with Pablo:
### (ServiceProvider, Operator, Verifier)

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -79,7 +79,7 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
      * @param caller The address of the caller.
```
tmigone, line 79:
```suggestion
     * @param serviceProvider The address of the serviceProvider.
```

pcarranzav:
> Hey @MoonBoi9001 @pcarranzav curious why you chose `(ServiceProvider, Operator, Verifier)` over `(ServiceProvider, Verifier, Operator)`? All the other staking functions follow the `(ServiceProvider, Verifier, ...)` pattern

I don't feel strongly about it, but to me it made intuitive sense that:
- Operator and service provider are closely related, so it made sense to have them close to each other
- It is coherent with other functions having (serviceProvider, verifier) because the ordering is transitive, i.e. (serviceProvider, operator), (operator, verifier), (serviceProvider, verifier) are valid orderings for functions that have only two of the three. And we already have some functions with (operator, verifier).

But if you think the other ordering is cleaner I'm happy with that too; but would prefer if the "transitiveness" was enforced, e.g. setOperator and setOperatorLocked should be changed to (verifier, operator, allowed) as well

tmigone:
I see. I think I still prefer to keep the (serviceProvider, verifier, ...) pattern, I feel like it's easier to remember, but I do agree we need to change SetOperator and SetOperatorLocked as well to enforce the transitiveness.

pcarranzav:
I'm happy with that, @MoonBoi9001 sorry for the churn, how do you feel about changing it to (serviceProvider, verifier, operator) and checking it's transitively consistent everywhere?

MoonBoi9001:
Okay, will do that.

pcarranzav (approved):
Looks great 👍

tmigone:
thanks @MoonBoi9001 and sorry for the trouble 👍

PR #1036, opened 2024-09-18
chore(SubgraphService): add current epoch to IndexingRewardsCollected…

Maikol:
… event


packages/subgraph-service/contracts/utilities/AllocationManager.sol
```diff
@@ -64,7 +64,8 @@ abstract contract AllocationManager is EIP712Upgradeable, GraphDirectory, Alloca
         uint256 tokensRewards,
         uint256 tokensIndexerRewards,
         uint256 tokensDelegationRewards,
-        bytes32 poi
+        bytes32 poi,
+        uint256 currentEpoch
```
tmigone, line 68: Nit: Can we add the param explanation to the comments?

tmigone (changes requested):

tmigone (approved):
🌋

PR #1041, opened 2024-09-18
fix: remove use of magic numbers in __ProvisionManager_init_unchained function (OZ N-11)

MoonBoi9001:
### Motivation:

- This PR addresses the [following N-11 OZ audit issue](https://defender.openzeppelin.com/#/audit/9d56f9fe-e25e-4ac5-acb0-772150889078/issues/N-11), applying the recommended fixes.

### Title: 
N-11 Use Constants for Default Values

### Details: 
In the ProvisionManager contract, the [__ProvisionManager_init_unchained function](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol#L151) initializes multiple default values such as type(uint256).min, type(uint256).max, type(uint32).min, type(uint32).max, type(uint64).min, and type(uint64).max directly within the function body.

This approach introduces magic numbers, reducing code readability and maintainability. Moreover, it could potentially introduce errors since the same literal values are used in [another part of the codebase](https://github.com/graphprotocol/contracts/blob/f47cf916c40b1d7aa10204d510ca1364c144736d/packages/subgraph-service/contracts/SubgraphService.sol#L436C5-L456C6) as well. In addition, the verifierCutRange is set between 0 and uint32.max but the maximum should be set to MAX_PPM.

Consider revising all uses of literal values, defining constants for them where applicable, and setting a more sensible maximum value for verifierCutRange.

##

### Key changes

- Removed the use of magic numbers inside the __ProvisionManager_init_unchained function, now using constants for default values.

### DIFF
--- a/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
+++ b/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
@@ -25,11 +25,11 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
 
     // Constants
     uint32 private constant DEFAULT_MIN_VERIFIER_CUT = type(uint32).min;
-    uint32 private constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
+    uint32 internal constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
     uint64 private constant DEFAULT_MIN_THAWING_PERIOD = type(uint64).min;
-    uint64 private constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
+    uint64 internal constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
     uint256 private constant DEFAULT_MIN_PROVISION_TOKENS = type(uint256).min;
-    uint256 private constant DEFAULT_MAX_PROVISION_TOKENS = type(uint256).max;
+    uint256 internal constant DEFAULT_MAX_PROVISION_TOKENS = type(uint256).max;
 
     /**
      * @notice Emitted when the provision tokens range is set.
--- a/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
+++ b/packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
@@ -24,11 +24,11 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
     using UintRange for uint256;
 
     // Constants
-    uint32 private constant DEFAULT_MIN_VERIFIER_CUT = type(uint32).min;
+    uint32 internal constant DEFAULT_MIN_VERIFIER_CUT = type(uint32).min;
     uint32 internal constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
-    uint64 private constant DEFAULT_MIN_THAWING_PERIOD = type(uint64).min;
+    uint64 internal constant DEFAULT_MIN_THAWING_PERIOD = type(uint64).min;
     uint64 internal constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
-    uint256 private constant DEFAULT_MIN_PROVISION_TOKENS = type(uint256).min;
+    uint256 internal constant DEFAULT_MIN_PROVISION_TOKENS = type(uint256).min;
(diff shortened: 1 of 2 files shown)


tmigone (changes requested):
We also need to use these variables in `SubgraphService.sol` in the following functions:
- `setMinimumProvisionTokens`
- `_getThawingPeriodRange`
- `_getVerifierCutRange`

That contract inherits from `ProvisionManager.sol` so the variables should be available.

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -23,6 +23,14 @@ import { ProvisionManagerV1Storage } from "./ProvisionManagerStorage.sol";
 abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionManagerV1Storage {
     using UintRange for uint256;
 
+    // Constants
+    uint32 private constant DEFAULT_MIN_VERIFIER_CUT = 0;
+    uint64 private constant DEFAULT_MIN_THAWING_PERIOD = 0;
+    uint256 private constant DEFAULT_MIN_PROVISION_TOKENS = 0;
+    uint64 private constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
+    uint256 private constant DEFAULT_MAX_PROVISION_TOKENS = type(uint256).max;
+    uint32 private constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
```
tmigone, line 32:
```suggestion
    uint32 private constant DEFAULT_MIN_VERIFIER_CUT = type(uint32).min;
    uint32 private constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
    uint64 private constant DEFAULT_MIN_THAWING_PERIOD = type(uint64).min;
    uint64 private constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
    uint256 private constant DEFAULT_MIN_PROVISION_TOKENS = type(uint256).min;
    uint256 private constant DEFAULT_MAX_PROVISION_TOKENS = type(uint256).max;
```

tmigone (changes requested):
Can we also use these vars in We also need to use these variables in `SubgraphService.sol` in the following functions:

- setMinimumProvisionTokens
- _getThawingPeriodRange
- _getVerifierCutRange

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -22,9 +22,14 @@ import { PPMMath } from "@graphprotocol/horizon/contracts/libraries/PPMMath.sol"
...
 import { LegacyAllocation } from "./libraries/LegacyAllocation.sol";
 
+import { ProvisionManager } from "@graphprotocol/horizon/contracts/data-service/utilities/ProvisionManager.sol";
+
+
+
 contract SubgraphService is
     Initializable,
     OwnableUpgradeable,
+    ProvisionManager,
```
tmigone, line 32: We shouldn't need to inherit `ProvisionManager` here as we are already inheriting `DataService` which in turn inherits `ProvisionManager` 😄

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -23,6 +23,14 @@ import { ProvisionManagerV1Storage } from "./ProvisionManagerStorage.sol";
 abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionManagerV1Storage {
     using UintRange for uint256;
 
+    // Constants
+    uint32 private constant DEFAULT_MIN_VERIFIER_CUT = type(uint32).min;
+    uint32 internal constant DEFAULT_MAX_VERIFIER_CUT = uint32(PPMMath.MAX_PPM);
+    uint64 private constant DEFAULT_MIN_THAWING_PERIOD = type(uint64).min;
+    uint64 internal constant DEFAULT_MAX_THAWING_PERIOD = type(uint64).max;
+    uint256 private constant DEFAULT_MIN_PROVISION_TOKENS = type(uint256).min;
+    uint256 internal constant DEFAULT_MAX_PROVISION_TOKENS = type(uint256).max;
```
tmigone, line 32: good call making these internal so they can be inherited.
I think we should make all internal, in this case the SubgraphService needed those 3 but other data services might need to use the rest so it's better if we make them available.

tmigone (changes requested):
Couple nits!

tmigone (approved):
💎

PR #1044, opened 2024-09-19
chore(SubgraphService): allow force closing over allocated allocations

Maikol:


packages/subgraph-service/contracts/utilities/AllocationManager.sol
```diff
@@ -475,4 +475,14 @@ abstract contract AllocationManager is EIP712Upgradeable, GraphDirectory, Alloca
...
+
+    /**
+     * @notice Checks if an allocation is over-allocated
+     * @param _indexer The address of the indexer
+     * @param _delegationRatio The delegation ratio to consider when locking tokens
+     * @return True if the allocation is over-allocated, false otherwise
+     */
+    function _isAllocationOverAllocated(address _indexer, uint32 _delegationRatio) internal view returns (bool) {
+        return !allocationProvisionTracker.check(_graphStaking(), _indexer, _delegationRatio);
+    }
```
tmigone, line 487:
```suggestion
    function _isOverAllocated(address _indexer, uint32 _delegationRatio) internal view returns (bool) {
        return !allocationProvisionTracker.check(_graphStaking(), _indexer, _delegationRatio);
    }
```

I would rename this function, it's the indexer that is over allocated not the allocation itself. 
Also I think it would be useful to expose this with an external function `function isOverAllocated(...) external ...` on the `SubgraphService` contract.

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -309,6 +309,18 @@ contract SubgraphService is
         _closeAllocation(allocationId);
     }
 
+    /**
+     * @notice See {ISubgraphService.closeOverAllocatedAllocation}
+     */
+    function closeOverAllocatedAllocation(address allocationId) external override {
+        Allocation.State memory allocation = allocations.get(allocationId);
+        require(
```
tmigone, line 317: We also need to check if the allocation is altruistic (tokens = 0). You shouldn't be able to close altruistic allos even if they are overallocated.

tmigone (changes requested):

packages/subgraph-service/contracts/SubgraphService.sol
Maikol: I thought about this but how can an altruistic allocation be over-allocated? If tokens == 0 then even if you have 0 available tokens you won't be over allocated. Am I missing something?

packages/subgraph-service/contracts/interfaces/ISubgraphService.sol
```diff
@@ -138,6 +144,20 @@ interface ISubgraphService is IDataServiceFees {
...
+     * This ensures rewards for other allocations are not diluted by an over-allocated allocation
+     *
+     * Requirements:
+     * - Allocation must exist and be open
+     * - Allocation must be over-allocated
+     *
+     * Emits a {AllocationClosed} event.
+     * @param allocationId The id of the allocation
+     */
+    function closeOverAllocatedAllocation(address allocationId) external;
```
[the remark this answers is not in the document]
Maikol: yeah, sounds good to me.

packages/subgraph-service/contracts/utilities/AllocationManager.sol
Maikol: Good idea, should we also include a `function isStaleAllocation()`?

tmigone: yes good catch!

packages/subgraph-service/contracts/SubgraphService.sol
Maikol: Right, yeah, makes sense, thank you.

packages/subgraph-service/contracts/interfaces/ISubgraphService.sol
```diff
@@ -112,36 +112,43 @@ interface ISubgraphService is IDataServiceFees {
...
      * @notice Thrown when trying to force close an altruistic allocation
      * @param allocationId The id of the allocation
      */
     error SubgraphServiceAllocationIsAltruistic(address allocationId);
 
+    /**
+     * @notice Thrown when trying to force close an allocation that is not over-allocated
+     * @param allocationId The id of the allocation
+     */
+    error SubgraphServiceAllocationNotOverAllocated(address allocationId);
```
tmigone, line 128: This one can be removed no?

packages/subgraph-service/contracts/interfaces/ISubgraphService.sol
```diff
@@ -112,36 +112,43 @@ interface ISubgraphService is IDataServiceFees {
     /**
      * @notice Thrown when trying to force close an allocation that is not stale
      * @param allocationId The id of the allocation
-     * @param lastPOIPresentedAt The timestamp when the last POI was presented
      */
-    error SubgraphServiceAllocationNotStale(address allocationId, uint256 lastPOIPresentedAt);
+    error SubgraphServiceCannotForceCloseAllocation(address allocationId);
```
tmigone, line 116: Here we need to update the comments to explain both cases where the error could be thrown (stale and overallocated)

tmigone (changes requested):
Couple nits

tmigone (approved):
🩵

PR #1046, opened 2024-09-20
chore(Horizon): reprovision only tokens thawed

Maikol:


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -693,13 +692,17 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
     /**
      * @notice See {IHorizonStakingMain-deprovision}.
      */
-    function _deprovision(address _serviceProvider, address _verifier, uint256 _nThawRequests) private {
+    function _deprovision(
+        address _serviceProvider,
+        address _verifier,
+        uint256 _nThawRequests
+    ) private returns (uint256 tokensThawed) {
```
tmigone, line 699:
```suggestion
    ) private returns (uint256) {
```

If we remove the variable name from the return value we should be able to keep the `tokensThawed` name on the code.

tmigone: Otherwise, we should be using underscore suffix to avoid name collisions (underscore prefix is reserved for non external variables like fn parameters).

tmigone (changes requested):
minor nit

packages/horizon/contracts/staking/HorizonStaking.sol
Maikol: I'll update to use the underscore as suffix, I think it helps with reading adding the parameter name to the return type.

tmigone (approved):
🏁

PR #1048, opened 2024-09-20
fix: tweaks to thawing pool management

tmigone:
A few notes:

**Provision**
- Thawing pool reset condition is `tokensThawing == 0` (empty pool)
- When a provision is fully slashed any ongoing thawing is cancelled
   - `sharesThawing` is set to 0
   - a nonce tied to the thawing pool is incremented, this invalidates existing thaw requests which in practice means when they are fulfilled they get 0 tokens.

**Delegation Pool**
- Delegation pool reset conditions are:
   - `tokens ==0 && shares ==0` (empty pool) or `tokens == thawingTokens` (pool fully thawing)
- When a delegation pool is fully slashed any ongoing thawing is cancelled
   - `sharesThawing` is set to 
   - nonce incremente to invalidate thaw requests
- Delegation pool invalid state
   - When the pool state is `tokens == 0 && shares != 0` as a result of a delegation pool being fully slashed
   - Pool can be recovered by using `addTokensToDelegationPool()`. This is the only delegation pool action that can be performed on invalid state pools
   - Delegators with shares thawing at the moment the pool gets to the invalid state loose their thawing shares and are not given delegation pool shares in return (they are all worth 0 tokens aynways)


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -286,7 +286,10 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
 
         // Delegation pool must exist before adding tokens
         DelegationPoolInternal storage pool = _getDelegationPool(serviceProvider, verifier);
-        require(pool.shares > 0, HorizonStakingInvalidDelegationPool(serviceProvider, verifier));
+        require(
+            pool.shares > 0 || pool.sharesThawing > 0,
```
Maikol, line 290: Why are we changing this? If `pool.tokens == pool.tokensThawing` won't we be adding tokens making it so next delegate won't initialize the pool again and be calculated with `pool.shares = 0`?

tmigone: My intention was to allow recovering an invalid pool when the entire shares were thawing. So `pool.shares = 0` and `pool.sharesThawing != 0`. But I later changed the slashing so that it forces `pool.sharesThawing = 0` if the pool is slashed 100%, so this scenario would never happen. I'll remove this change

tmigone: good catch btw

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -666,25 +682,40 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
...
+        // Calculate shares to issue
+        // Thawing pool is reset/initialized in any of the following cases:
+        // - prov.tokensThawing == 0, pool is empty
+        bool initializeThawingPool = prov.tokensThawing == 0;
+        uint256 thawingShares = initializeThawingPool ? _tokens : ((prov.sharesThawing * _tokens) / prov.tokensThawing);
         uint64 thawingUntil = uint64(block.timestamp + uint256(prov.thawingPeriod));
 
-        prov.tokensThawing = prov.tokensThawing + _tokens;
+        // Safety check to ensure pool has no shares when being initialized
+        // This should never happen if the pool was reset due to slashing or it's the first initialization
```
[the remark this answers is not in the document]
tmigone: Nah i think ill just remove that safety check. Only way `tokensThawing` can be zero are:
- when emptying the thawing pool by withdrawing funds, which should also burn the `sharesThawing` and send it back to 0.
- when the provision is fully slashed, which should already set `sharesThawing` to 0.

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -923,20 +993,30 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
...
+        // process - only fulfill thaw requests for the current valid nonce
+        uint256 tokens = 0;
+        bool invalidThawRequest = thawRequest.thawingNonce != thawingNonce;
+        if (!invalidThawRequest) {
+            tokens = (thawRequest.shares * tokensThawing) / sharesThawing;
+            tokensThawing = tokensThawing - tokens;
+            sharesThawing = sharesThawing - thawRequest.shares;
+            tokensThawed = tokensThawed + tokens;
+        }
+        emit ThawRequestFulfilled(
```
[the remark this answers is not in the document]
tmigone: I thought about it, but I think request is fulfilled either way, we are not skipping it, it's just that it's fulfilled and the result is 0.

packages/horizon/test/staking/delegation/addToPool.t.sol
```diff
@@ -86,5 +86,62 @@ contract HorizonStakingDelegationAddToPoolTest is HorizonStakingTest {
...
+        uint256 recoverAmount
+    ) public useIndexer useProvision(tokens, 0, 0) useDelegationSlashing() {
+        recoverAmount = bound(recoverAmount, 1, MAX_STAKING_TOKENS);
+        delegationTokens = bound(delegationTokens, MIN_DELEGATION, MAX_STAKING_TOKENS);
+
+        // create delegation pool
+        resetPrank(users.delegator);
+        _delegate(users.indexer, subgraphDataServiceAddress, delegationTokens, 0);
+
+        // undelegate half shares so we have some thawing shares/tokens
```
Maikol, line 124: very minor but this comment is incorrect I believe, the test is undelegating all shares

Maikol (approved):
Looks good to me 🥇

packages/horizon/test/staking/delegation/addToPool.t.sol
tmigone: amazing, don't know how you spotted that lol

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -429,6 +437,16 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
                     (pool.tokensThawing * (FIXED_POINT_PRECISION - delegationFractionSlashed)) /
                     FIXED_POINT_PRECISION;
 
+                // If the delegation pool was fully slashed pending thawings are cancelled by:
+                // - deleting all thawing shares
+                // - incrementing the nonce to invalidate pending thaw requests
+                // Note that thawing shares are completely lost, delegators won't get back the corresponding
+                // delegation pool shares.
+                if (pool.tokens == 0 && pool.tokensThawing == 0) {
```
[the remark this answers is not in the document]
pcarranzav: Thinking if we didn't slash the whole pool but for rounding issues we slashed the whole thawing pool we should increment the nonce

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -803,7 +850,12 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
         uint256 _nThawRequests
     ) private {
         DelegationPoolInternal storage pool = _getDelegationPool(_serviceProvider, _verifier);
-        require(pool.tokens != 0, HorizonStakingInvalidDelegationPoolState(_serviceProvider, _verifier));
+
+        // An invalid delegation pool has shares but no tokens
+        require(
```
pcarranzav, line 855: Not sure if this would behave well with `pool.shares == 0` and `pool.tokens == 0`, see below we do `pool.tokens = pool.tokens - tokensThawed`. So I think we should always enforce `pool.tokens != 0` here?

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -923,20 +985,30 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
...
-        // process
-        uint256 tokens = (thawRequest.shares * tokensThawing) / sharesThawing;
-        tokensThawing = tokensThawing - tokens;
-        sharesThawing = sharesThawing - thawRequest.shares;
-        tokensThawed = tokensThawed + tokens;
-        emit ThawRequestFulfilled(_thawRequestId, tokens, thawRequest.shares, thawRequest.thawingUntil);
+        // process - only fulfill thaw requests for the current valid nonce
+        uint256 tokens = 0;
+        bool invalidThawRequest = thawRequest.thawingNonce != thawingNonce;
+        if (!invalidThawRequest) {
```
pcarranzav, line 996: micronit: we're always using`!invalidThawRequest`, so might be slightly cleaner to use a `validThawRequest` instead?

packages/horizon/contracts/staking/HorizonStaking.sol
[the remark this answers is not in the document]
tmigone: Rounding issues might be a problem agreed. The thing is removing the `pool.tokens == 0` means the nonce would get incremented if there is a partial slashing and an empty thawing pool, which is not really a problem but it's also not necessary.  

Perhaps we can remove the `pool.tokens == 0` check but also check that the thawing pool was actually slashed down to zero (as opposed to already being zero before the slashing). Wdyt?

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -402,6 +402,14 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
                 (FIXED_POINT_PRECISION);
             prov.tokens = prov.tokens - providerTokensSlashed;
 
+            // If the provision was fully slashed pending thawings are cancelled by:
+            // - deleting all thawing shares
+            // - incrementing the nonce to invalidate pending thaw requests
+            if (prov.tokens == 0 && prov.tokensThawing == 0) {
```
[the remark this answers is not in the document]
tmigone: Same as below.

packages/horizon/contracts/staking/HorizonStaking.sol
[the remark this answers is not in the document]
pcarranzav: I guess the problem arises when `tokensThawing == 0 && pool.sharesThawing != 0`, so maybe we could check for that condition?

packages/horizon/contracts/staking/HorizonStaking.sol
tmigone: I guess the condition should be `pool.tokens >= tokensThawed`, no?

Theoretically it should always be true because `pool.tokens >= pool.tokensThawing` and `pool.tokensThawing >= tokensThawed`. The only scenario where it gets tricky is when the thawing pool is reset due to the delegation pool being fully slashed, but in that case the nonce invalidation ensures that `tokensThawed = 0`. 

Do you still think we should add a check here?

packages/horizon/contracts/staking/HorizonStaking.sol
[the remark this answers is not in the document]
tmigone: yep that's even better, like it!

packages/horizon/contracts/staking/HorizonStaking.sol
pcarranzav: I see, so this would allow clearing up those thaw requests... that makes sense, maybe worth adding a comment here noting we might have a case where `shares == 0` and that then the tokensThawed should be 0?

Maikol (approved):

pcarranzav (approved):
LGTM 🎸

PR #1056, opened 2024-10-04
feat: add rav collected event

tmigone:


Maikol (approved):
I think we currently don't have tests for this, right? 😅

tmigone:
We have some on the subgraph service but yeah we should have full coverage here as well 😃

PR #1060, opened 2024-10-10
chore(Horizon): add signers to TAPCollector

Maikol:


packages/horizon/contracts/interfaces/ITAPCollector.sol
```diff
@@ -34,6 +43,36 @@ interface ITAPCollector is IPaymentsCollector {
         bytes signature;
     }
 
+    /**
+     * @notice Emitted when a signer is authorized to sign RAVs for a payer
+     * @param authorizedSigner The address of the authorized signer
+     * @param payer The address of the payer authorizing the signer
+     */
+    event AuthorizeSigner(address indexed authorizedSigner, address indexed payer);
```
pcarranzav, line 51: nit: I see you've used the same event signatures as in TAP verifier, but I wonder if we could use the opportunity to make them consistent with other events in two ways:
1) Ensure order is consistent, e.g. always have payer, authorizedSigner, thawEndTimestamp in that order
2) Use more descriptive verb tenses, e.g. SignerAuthorized, SignerThawing, SignerThawingCanceled, SignerRevoked

packages/horizon/contracts/interfaces/ITAPCollector.sol
```diff
@@ -54,6 +93,50 @@ interface ITAPCollector is IPaymentsCollector {
         bytes signature
     );
 
+    /**
+     * Thrown when the signer is already authorized
+     * @param signer The address of the signer
+     * @param authorizingPayer The address of the payer authorizing the signer
+     */
+    error TAPCollectorSignerAlreadyAuthorized(address signer, address authorizingPayer);
```
pcarranzav, line 101: nit: same as with events, it'd be good to standardize on the ordering: payer, signer

packages/horizon/contracts/payments/collectors/TAPCollector.sol
```diff
@@ -29,21 +30,89 @@ contract TAPCollector is EIP712, GraphDirectory, ITAPCollector {
...
      * @param eip712Version The version of the EIP712 domain.
      * @param controller The address of the Graph controller.
+     * @param _revokeSignerThawingPeriod The duration (in seconds) in which a signer is thawing before they can be revoked.
      */
     constructor(
         string memory eip712Name,
         string memory eip712Version,
-        address controller
-    ) EIP712(eip712Name, eip712Version) GraphDirectory(controller) {}
+        address controller,
```
pcarranzav, line 53: meganit: can we standardize on using underscores for function parameters everywhere, or not using them anywhere? I think in other contracts we're not using it, so here we could use a slightly different name rather than adding an underscore?

packages/horizon/contracts/payments/collectors/TAPCollector.sol
```diff
@@ -29,21 +30,89 @@ contract TAPCollector is EIP712, GraphDirectory, ITAPCollector {
...
+    /**
+     * See {ITAPCollector.authorizeSigner}.
+     */
+    function authorizeSigner(address signer, uint256 proofDeadline, bytes calldata proof) external override {
+        require(
+            authorizedSigners[signer].payer == address(0),
+            TAPCollectorSignerAlreadyAuthorized(signer, authorizedSigners[signer].payer)
+        );
+
+        verifyAuthorizedSignerProof(proof, proofDeadline, signer);
```
pcarranzav, line 68: nit: can we call it `_verifyAuthorizedSignerProof` to make it clearer it's a private func?

packages/horizon/contracts/payments/collectors/TAPCollector.sol
```diff
@@ -137,4 +173,74 @@ contract TAPCollector is EIP712, GraphDirectory, ITAPCollector {
...
+        if (tokensToCollect > 0) {
+            _graphPaymentsEscrow().collect(
+                paymentType,
+                payer,
+                receiver,
+                tokensToCollect,
+                dataService,
+                tokensDataService
+            );
+            tokensCollected[dataService][receiver][payer] = tokensRAV;
```
pcarranzav, line 231: The call to `_graphPaymentsEscrow().collect` is to a trusted contract so I don't see a direct risk of reentrancy, but it'd  still be cleaner/safer to move this before that call

packages/horizon/test/payments/tap-collector/collect.t.sol
```diff
@@ -0,0 +1,90 @@
+// SPDX-License-Identifier: MIT
```
pcarranzav, line 1: It'd be nice to add a test trying to collect when the signer isn't authorized

pcarranzav (approved):
LGTM (one testing detail but can be fixed separately)

PR #1062, opened 2024-10-17
fix: added missing parameters to deploy script and reduced runs to 50 for SubgraphService

Maikol:
Run gas profile for both options and saving for 200 runs was:

```
SubgraphService:
  acceptProvisionPendingParameters: 693 gas saved
  collect: 1,405 gas saved
  initialize: 991 gas saved
  startService: 914 gas saved
  stopService: 295 gas saved

DisputeManager:
  acceptDispute: 248 gas saved
  cancelDispute: 66 gas saved
  createIndexingDispute: 469 gas saved
  createQueryDispute: 481 gas saved
  initialize: 387 gas saved
```


packages/subgraph-service/hardhat.config.ts
```diff
@@ -13,7 +13,7 @@ const config: HardhatUserConfig = {
     settings: {
       optimizer: {
         enabled: true,
-        runs: 200,
+        runs: 50,
```
matiasedgeandnode, line 16: This is also set in `graphprotocol/contracts/packages/subgraph-service/foundry.toml`. Not sure if they are meant to be kept in-sync...

Maikol: ah yes, thank you! I changed it back to 200 to get the gas report and forgot to switch it back

tmigone (approved):

PR #1067, opened 2024-11-05
ci: fix e2e-contracts.yml

matiasedgeandnode:
I would rather update `nitro-testnode` but not sure why it's pinned to `c47cb8c643bc8e63ff096f7f88f9152064d1532a`. Since the `e2e` tests are failing, can't really blindly update that version.


tmigone:
So I had to pin it because offchainlabs was pretty aggressive when making changes to the repo at some point, breaking changes where you wouldn't expect it.

That said, at this point I believe we no longer need the e2e tests at least not in the current form. Originally they were meant to run on a pair of L1/L2 chains where we deployed the protocol on both, deployed the bridge, configured it and then ran some cross-chain actions. Now with the L1 protocol deprecated plus almost 2 years of the bridge running fine I don't see the value in keeping these e2e tests, specially considering how expensive (in computation) they are and how often they fail.

What I think is much more interesting is for us to be able to run a test suite against a fork from arbitrum one for example. Perhaps using tenderly or similar. Though maybe horizon is the time and place to do this!

Thoughts? @matiasedgeandnode @pcarranzav

matiasedgeandnode:
I don't have enough context here, but would be very happy if we delete tests that are irrelevant, especially if they are flaky.

tmigone:
Agreed, let's remove the `e2e-contracts` workflow so they don't run anymore on the CI. We can keep the e2e test files though in the repo because they can be useful if we wanted to run those locally.

matiasedgeandnode:
> Agreed, let's remove the `e2e-contracts` workflow so they don't run anymore on the CI. We can keep the e2e test files though in the repo because they can be useful if we wanted to run those locally. With Horizon we'll probably clean this up a lot.

done

tmigone (approved):

PR #1072, opened 2024-11-26
TRST audit fixes for Graph Payments contracts

tmigone:
This PR addresses the following audit findings:
- `TRST-H-1 A payer could bypass the escrow mechanism and avoid payments`
- `TRST-M-10 A RAV could be collected more than once, leading to double payment`
- `TRST-CL-1 A payer could bypass the escrow mechanism and avoid payments through
the vulnerable collector allowance mapping`
- `TTRST-L-10 The getBalance() function could revert when balance is lower than tokens
thawing`
- `TRST-M-5 Lack of chunking functionality of new RAVs may cause them to not be
processable`
- `TRST-L-12 The collection cuts could exceed 100% causing collect() to revert`


Maikol (approved):

packages/horizon/test/payments/tap-collector/collect/collect.t.sol
```diff
@@ -49,7 +49,7 @@ contract TAPCollectorCollectTest is TAPCollectorTest {
      * TESTS
      */
 
-    function testTAPCollector_Collect(
+    function testTAPCollector_Collect_Only(
```
Maikol, line 52: Was this to run only this test?

packages/subgraph-service/test/subgraphService/collect/query/query.t.sol
```diff
@@ -80,7 +80,7 @@ contract SubgraphServiceRegisterTest is SubgraphServiceTest {
      * TESTS
      */
 
-    function testCollect_QueryFees(
+    function testCollect_QueryFees_Only(
```
Maikol, line 83: Same question about only here.

Maikol (approved):
Looks good!

Maikol (approved):

PR #1073, opened 2024-11-28
TRST audit fixes for HorizonStaking contract

Maikol:
This PR fixes the following issues:
- `TRST-H-2 Core thawing logic of Horizon staking can be broken due to collision in data
structures`
- `TRST-H-3 An attacker could prevent a delegator from ever withdrawing their tokens`
- `TRST-H-4 The last withdrawal from a provision or delegation could permanently revert
due to accounting flaw in slashing`
- `TRST-H-6 During the transition period, legacy allocations cannot be slashed`
- `TRST-H-7 Delegators that began undelegation before the transition would not be able
to withdraw them`
- `TRST-M-3 An attacker could make the minimum delegation amount arbitrarily large
and prevent competing delegations`
- `TRST-M-11 After the transition period, locked amount would still not be available for
use`
- `TRST-M-12 The operator check in closeAllocations() will not work, indexers must close
by themselves`
- `TRST-L-3 The getThawedTokens() function could return a wrong amount leading to
integration risks`
- `TRST-R-1 Improve event structure`
- `TRST-R-7 Thawing shares should be rounded up to protect from early unlocking of
tokens`
- `TRST-R-9 Documentation errors`

### DIFF
--- a/packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
+++ b/packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
@@ -206,6 +206,15 @@ interface IHorizonStakingMain {
     );
 
     /**
+     * @notice Emitted when `delegator` withdrew delegated `tokens` from `indexer` using `withdrawDelegated`.
+     * @dev This event is for the legacy `withdrawDelegated` function.
+     * @param indexer The address of the indexer
+     * @param delegator The address of the delegator
+     * @param tokens The amount of tokens withdrawn
+     */
+    event StakeDelegatedWithdrawn(address indexed indexer, address indexed delegator, uint256 tokens);
+
+    /**
      * @notice Emitted when tokens are added to a delegation pool's reserve.
      * @param serviceProvider The address of the service provider
      * @param verifier The address of the verifier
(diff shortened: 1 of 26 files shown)


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -842,6 +845,7 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
      * have been done before calling this function.
      */
     function _delegate(address _serviceProvider, address _verifier, uint256 _tokens, uint256 _minSharesOut) private {
+        require(_tokens >= MIN_DELEGATION, HorizonStakingInsufficientTokens(_tokens, MIN_DELEGATION));
```
tmigone, line 848: Can we document why we need a minimum delegation threshold?

tmigone, line 848: I think it's worth creating a new error, this one is already being thrown in several places.

tmigone (approved):
Some suggestions!

packages/horizon/contracts/staking/HorizonStakingExtension.sol
```diff
@@ -255,6 +263,92 @@ contract HorizonStakingExtension is HorizonStakingBase, IHorizonStakingExtension
...
+        _graphToken().pushTokens(beneficiary, reward);
+
+        emit StakeSlashed(indexer, tokens, reward, beneficiary);
+    }
+
+    /**
+     * @notice Withdraw undelegated tokens once the unbonding period has passed.
+     * @param indexer Withdraw available tokens delegated to indexer
+     */
+    function legacyWithdrawDelegated(
```
pcarranzav, line 319: Maybe including this in withdrawDelegated() (as suggested in the audit report) would make it simpler for Explorer? otherwise Explorer might need to do some tricky handling to figure out what function to call.

Maikol: I think the explorer will still need to do some handling to know which type of withdraw it needs to use, right? Once Horizon is deployed undelegating will use the new thaw requests so it'll need to call the new withdraw. But also now that I think about it we won't be able to remove this functionality since we'll still need to support withdrawing after transition period I guess so I'll move it to `withdrawDelegated()`.

Maikol: Moved. Do you mind reviewing again? Thanks.

packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
```diff
@@ -433,6 +433,13 @@ interface IHorizonStakingMain {
      */
     error HorizonStakingInsufficientDelegationTokens(uint256 tokens, uint256 minTokens);
 
+    /**
+     * @notice Thrown when the minimum token amount required for delegation with beneficiary is not met.
```
tmigone, line 437: delegation -> undelegation

tmigone (approved):
🔥

PR #1085, opened 2025-01-09
partial: indexing payments [DO NOT MERGE]

matiasedgeandnode:
Adds Authorizable and DIPCollector.

PR #1089, opened 2025-01-16
fix: cleanup TAPCollectorAuthorizationAlreadyRevoked

matiasedgeandnode:
This PR simply cleans up what's already there, but if you would consider a slightly bigger re-write, maybe the collector would benefit from a `modifier onlyAuthorized(address _signer)` for `thawSigner`, `cancelThawSigner` and `revokeAuthorizedSigner` like what's proposed by #1085 -> [here](https://github.com/graphprotocol/contracts/pull/1085/files#diff-d5870b8cb11b87807ea66ca4948b55b9c847d3ad74734a25331574ce54be011dR27).


tmigone:
I like what you did with Authorizable! Want to PR it into `horizon`?

PR #1092, opened 2025-01-24
fix: Fix incorrect path for solhint-graph-config Update README.md

mdqst:
I noticed a typo in the packages table where the path for `solhint-graph-config` was incorrectly pointing to `./packages/eslint-graph-config`.
I’ve updated it to the correct path: `./packages/solhint-graph-config`.
This ensures the documentation accurately reflects the package location.


tmigone:
ty!

PR #1096, opened 2025-01-28
chore: reduce contract sizes to fit 24kb HorizonStaking and SubgraphService

Maikol:


packages/subgraph-service/contracts/utilities/AllocationManagerStorage.sol
```diff
@@ -6,7 +6,7 @@ import { LegacyAllocation } from "../libraries/LegacyAllocation.sol";
 
 abstract contract AllocationManagerV1Storage {
     /// @notice Allocation details
-    mapping(address allocationId => Allocation.State allocation) public allocations;
+    mapping(address allocationId => Allocation.State allocation) internal allocations;
 
     /// @notice Legacy allocation details
     mapping(address allocationId => LegacyAllocation.State allocation) public legacyAllocations;
```
tmigone, line 12: Can we make this internal as well for consistency? (there is a getter)

packages/subgraph-service/contracts/utilities/AllocationManagerStorage.sol
```diff
@@ -15,14 +15,14 @@ abstract contract AllocationManagerV1Storage {
     mapping(address indexer => uint256 tokens) public allocationProvisionTracker;
 
     /// @notice Maximum amount of time, in seconds, allowed between presenting POIs to qualify for indexing rewards
-    uint256 public maxPOIStaleness;
+    uint256 internal maxPOIStaleness;
```
tmigone, line 18: this could be made public if we make `legacyAllocations` internal? Also i dont think this value is accesible otherwise no?

PR #1097, opened 2025-01-28
chore: add collectorId to TAPCollector tokens tracking

Maikol:


packages/horizon/contracts/interfaces/ITAPCollector.sol
```diff
@@ -25,6 +25,8 @@ interface ITAPCollector is IPaymentsCollector {
 
     /// @notice The Receipt Aggregate Voucher (RAV) struct
     struct ReceiptAggregateVoucher {
+        // The ID of the payment
+        address paymentId;
```
tmigone, line 29: Thinking about it a bit more I think the term `collectionId` is a bit more representative of what this is. For each data service these are "collection buckets", wdyt?

tmigone, line 29:
```suggestion
        bytes32 paymentId;
```

`bytes32` feels a better choice here.

tmigone (changes requested):
We also need to update `_collectQueryFees` in `SubgraphService.sol`, currently it uses RAV metadata to encode the `allocationId`, which will now be found as a field in the RAV.

tmigone (approved):

PR #1100, opened 2025-02-06
fix: Horizon migration script fixes

Maikol:


packages/horizon/hardhat.config.ts
```diff
@@ -1,15 +1,30 @@
...
   require('hardhat-graph-protocol')
 }
 
-export default hardhatBaseConfig
+const config: HardhatUserConfig = {
+  ...hardhatBaseConfig,
+  graph: {
+    deployments: {
+      ...hardhatBaseConfig.graph?.deployments,
+      subgraphService: false
```
tmigone, line 21: I think the cleanest approach is to only have horizon deployment in the base config, then each project would add whatever is of interest to them. Horizon is the common denominator for all data services after all.

tmigone (approved):

PR #1109, opened 2025-02-12
Horizon: Integration tests

Maikol:


packages/horizon/tasks/transitions.ts
```diff
@@ -0,0 +1,52 @@
+import { task, types } from 'hardhat/config'
+import { createBanner } from '../utils/banners'
+import { ethers } from 'ethers'
+
+task('transition:unset-subgraph-service', 'Unsets the subgraph service in HorizonStaking')
+  .addOptionalParam('governorIndex', 'Index of the governor account in getSigners array', 0, types.int)
+  .setAction(async (taskArgs, hre) => {
+    console.log(createBanner('UNSETTING SUBGRAPH SERVICE'))
```
tmigone, line 8: Should we add:


```
  // Check that we're on a local network
  if (hre.network.name !== 'localhost' && hre.network.name !== 'hardhat') {
    throw new Error('This task can only be run on localhost or hardhat network')
  }
```

tmigone: to all tasks in this file as well

packages/horizon/tasks/transitions.ts
```diff
@@ -0,0 +1,52 @@
+import { task, types } from 'hardhat/config'
+import { createBanner } from '../utils/banners'
+import { ethers } from 'ethers'
+
+task('transition:unset-subgraph-service', 'Unsets the subgraph service in HorizonStaking')
+  .addOptionalParam('governorIndex', 'Index of the governor account in getSigners array', 0, types.int)
```
tmigone, line 6:
```suggestion
  .addOptionalParam('governorIndex', 'Derivation path index for the governor account', 0, types.int)
```

packages/horizon/tasks/transitions.ts
```diff
@@ -0,0 +1,52 @@
...
+    const rewardsManager = hre.graph().horizon!.contracts.RewardsManager
+
+    console.log('Unsetting subgraph service...')
+    const tx = await rewardsManager.connect(governor).setSubgraphService(ethers.ZeroAddress)
+    await tx.wait()
+    console.log('Subgraph service unset')
+  })
+
+task('transition:clear-thawing', 'Clears the thawing period in HorizonStaking')
+  .addOptionalParam('governorIndex', 'Index of the governor account in getSigners array', 0, types.int)
```
tmigone, line 22:
```suggestion
  .addOptionalParam('governorIndex', 'Derivation path index for the governor account', 0, types.int)
```

packages/horizon/tasks/transitions.ts
```diff
@@ -0,0 +1,52 @@
...
+    const horizonStaking = hre.graph().horizon!.contracts.HorizonStaking
+
+    console.log('Clearing thawing period...')
+    const tx = await horizonStaking.connect(governor).clearThawingPeriod()
+    await tx.wait()
+    console.log('Thawing period cleared')
+  })
+
+task('transition:enable-delegation-slashing', 'Enables delegation slashing in HorizonStaking')
+  .addOptionalParam('governorIndex', 'Index of the governor account in getSigners array', 0, types.int)
```
tmigone, line 37:
```suggestion
  .addOptionalParam('governorIndex', 'Derivation path index for the governor account', 0, types.int)
```

packages/horizon/scripts/e2e/pre-upgrade.ts
```diff
@@ -0,0 +1,197 @@
...
+  const stakingAddress = arbSepoliaAddresses.L2Staking.address
+  const graphTokenAddress = arbSepoliaAddresses.L2GraphToken.address
+
+  console.log(`Using Staking contract at: ${stakingAddress}`)
+  console.log(`Using GraphToken contract at: ${graphTokenAddress}`)
+
+  // Create contract instances
+  const provider = new HardhatEthersProvider(hre.network.provider, hre.network.name)
+  const GraphToken = new Contract(graphTokenAddress, graphTokenABI, provider) as unknown as IGraphToken
+  const Staking = new Contract(stakingAddress, combinedStakingABI, provider) as unknown as IStaking
```
tmigone, line 60: again not sure why we are manually instancing contracts and providers instead of using hre.graph()

packages/horizon/scripts/e2e/pre-upgrade.ts
```diff
@@ -0,0 +1,197 @@
...
+  console.log(`Asset holder balance: ${assetHolderBalance}`)
+
+  // Convert BigNumber to bigint for comparison
+  if (assetHolderBalance < ethers.parseEther('20000000')) {
+    throw new Error('Asset holder balance is less than 20M tokens')
+  }
+
+  // Impersonate the account
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const grtHolder = await ethers.getImpersonatedSigner(GRT_HOLDER_ADDRESS) as any
```
tmigone, line 73: How about instead of relying on a known account with big chunk of GRT we use the governor to mint some?

packages/horizon/scripts/e2e/e2e
```diff
@@ -0,0 +1,115 @@
...
+npx hardhat deploy:migrate --network localhost --horizon-config e2e-test --step 2 --patch-config --signer-index 1 --hide-banner
+
+# Step 3 - Deployer
+npx hardhat deploy:migrate --network localhost --horizon-config e2e-test --step 3 --patch-config --signer-index 0 --hide-banner
+
+# Step 4 - Governor
+npx hardhat deploy:migrate --network localhost --horizon-config e2e-test --step 4 --patch-config --signer-index 1 --hide-banner
+
+# Unset subgraph service
+npx hardhat transition:unset-subgraph-service --network localhost --governor-index 1
```
tmigone, line 96: Why are we unsetting the subgraph service? Aren't horizon deployment scripts already setting it to 0x00? (I have to fix that yet).

packages/horizon/test/integration/during-transition-period/service-provider.test.ts
```diff
@@ -0,0 +1,376 @@
...
+  let graphToken: IGraphToken
+  let snapshotId: string
+
+  // TODO: FIX THIS
+  const subgraphServiceAddress = '0x254dffcd3277C0b1660F6d42EFbB754edaBAbC2B'
+
+  before(() => {
+    const graph = hre.graph()
+
+    horizonStaking = graph.horizon!.contracts.HorizonStaking as unknown as IHorizonStaking
```
tmigone, line 26: ugh i need to fix this.

packages/horizon/test/integration/during-transition-period/service-provider.test.ts
```diff
@@ -0,0 +1,376 @@
...
+import { indexers } from '../../../scripts/e2e/fixtures/indexers'
+
+describe('Service Provider', () => {
+  let horizonStaking: IHorizonStaking
+  let rewardsManager: IRewardsManager
+  let graphToken: IGraphToken
+  let snapshotId: string
+
+  // TODO: FIX THIS
+  const subgraphServiceAddress = '0x254dffcd3277C0b1660F6d42EFbB754edaBAbC2B'
```
tmigone, line 21: Im confused, these tests run against a deployment where the subgraph service does not exist, what am i missing?

packages/hardhat-graph-protocol/src/sdk/hardhat.base.config.ts
```diff
@@ -60,9 +60,12 @@ export const networksUserConfig: BaseNetworksUserConfig = {
     secureAccounts: {
       enabled: true,
     },
-    ...(vars.has('LOCALHOST_ACCOUNTS_MNEMONIC') && {
-      accounts: { mnemonic: vars.get('LOCALHOST_ACCOUNTS_MNEMONIC') },
-    }),
+    ...(vars.has('FORK') && vars.get('FORK') === 'true'
```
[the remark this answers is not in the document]
Maikol: Without setting `accounts: 'remote'` I wasn't able to impersonate accounts correctly. Changing this made it possible.

packages/horizon/test/integration/after-transition-period/delegator.test.ts
tmigone, line 1: A few more cases:
- Delegator that delegated before transition period (existing user)
   - Undelegates after transtition period and withdraws after transition period
   - Undelegates before transition period and withdraws after transition period

packages/horizon/hardhat.config.ts
```diff
@@ -1,13 +1,28 @@
...
+
+// Hardhat tasks
+function loadTasks() {
+  const tasksPath = join(__dirname, 'tasks')
+  readdirSync(tasksPath)
+    .filter(pth => pth.includes('.ts'))
+    .forEach((file) => {
+      require(join(tasksPath, file))
+    })
+}
```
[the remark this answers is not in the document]
Maikol: I believe we can't use `hre` inside `hardhat.config.ts`. I tried it with `__dirname` and it seems to work. What do you think?

packages/horizon/hardhat.config.ts
```diff
@@ -1,13 +1,28 @@
...
+
+if (existsSync(join(__dirname, 'build/contracts'))) {
```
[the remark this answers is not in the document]
Maikol: Same as above, can't use `hre` here.

packages/horizon/test/integration/shared/staking.ts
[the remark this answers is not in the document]
tmigone: btw when i say SDK i mean `packages/hardhat-graph-protocol/src/sdk`

packages/horizon/test/integration/during-transition-period/slasher.test.ts
```diff
@@ -0,0 +1,104 @@
...
+    it('should be able to slash indexer stake', async () => {
+      // Before slash state
+      const idleStakeBeforeSlash = await horizonStaking.getIdleStake(indexer)
+      const tokensVerifier = tokensToSlash / 2n
+      const slasherBeforeBalance = await graphToken.balanceOf(slasher.address)
+
+      // Slash tokens
+      await slash({
+        horizonStaking,
+        verifier: slasher,
```
tmigone, line 56: This is not legacy slashing no?

packages/horizon/test/integration/during-transition-period/permissionless.test.ts
```diff
@@ -0,0 +1,72 @@
...
+      // Get signers
+      indexer = await ethers.getSigner(indexers[0].address)
+      anySigner = (await ethers.getSigners())[19]
+
+      // Get allocation details
+      allocationID = indexers[0].allocations[0].allocationID
+      allocationTokens = indexers[0].allocations[0].tokens
+    })
+
+    it('should allow any user to close an allocation with zero POI', async () => {
```
tmigone, line 52:
```suggestion
    it('should allow any user to close an allocation with zero POI after 28 epochs', async () => {
```

packages/horizon/test/integration/during-transition-period/service-provider.test.ts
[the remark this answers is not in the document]
tmigone: Adding more notes to myself mostly, don't worry about this for this PR:
- Delegation parameter config multicall tx

packages/horizon/hardhat.config.ts
```diff
@@ -1,13 +1,28 @@
...
+  loadTasks()
+}
 
 // Skip importing hardhat-graph-protocol when building the project, it has circular dependency
 if (process.env.BUILD_RUN !== 'true') {
```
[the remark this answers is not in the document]
Maikol: 1. True, I'll remove it.
2. Yeah, thinking about it maybe we should have this and `loadTasks()` together? Though not if we move load tasks to the sdk.

packages/horizon/package.json
```diff
@@ -16,7 +16,8 @@
     "lint": "yarn lint:ts && yarn lint:sol",
     "clean": "rm -rf build dist cache cache_forge typechain-types",
     "build": "BUILD_RUN=true hardhat compile",
-    "test": "forge test && hardhat test"
+    "test": "forge test",
+    "test:e2e": "./scripts/e2e/e2e"
```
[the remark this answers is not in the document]
Maikol: Of course.

packages/horizon/hardhat.config.ts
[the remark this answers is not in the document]
tmigone: ohhhhh trueee. let's stick with what we have then!

packages/horizon/hardhat.config.ts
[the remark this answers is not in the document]
tmigone: same as other, lets keep it then

packages/horizon/scripts/e2e/transfer-ownership.ts
```diff
@@ -0,0 +1,98 @@
...
+  // Get contract addresses
+  const addressesJson = require('@graphprotocol/contracts/addresses.json')
+  const arbSepoliaAddresses = addressesJson['421614']
+  const stakingAddress = arbSepoliaAddresses.L2Staking.address
+
+  // Get ABIs
+  const combinedStakingABI = mergeABIs(L2StakingABI, StakingExtensionABI)
+
+  // Get contracts
+  const staking = new Contract(stakingAddress, combinedStakingABI, ethers.provider) as unknown as IStaking
```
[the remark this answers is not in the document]
Maikol: We run this script before running the upgrade so we won't have the horizon contracts deployed yet. I guess we might be able to use `graph().horizon!` this but it might be possible that the addresses on `addresses-local.json` are different.

packages/horizon/scripts/e2e/transfer-ownership.ts
```diff
@@ -0,0 +1,98 @@
...
+  await acceptProxyAdminTx.wait()
+  console.log(`New GraphProxyAdmin governor: ${await graphProxyAdmin.governor()}`)
+
+  console.log('\n--- STEP 3: Assign new slasher ---')
+
+  // Assign new slasher
+  console.log('Assigning new slasher...')
+  const slasherTx = await staking.connect(newGovernor).setSlasher(newSlasher.address, true)
+  await slasherTx.wait()
+  console.log(`New slasher: ${newSlasher.address}, allowed: ${await staking.slashers(newSlasher.address)}`)
```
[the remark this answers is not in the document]
Maikol: Yes. We could also impersonate the `DisputeManager` account but I thought adding the slasher would avoid having impersonate on the test.

packages/horizon/scripts/e2e/e2e
Maikol: This is just temporary since currently the migrate script in horizon requires the subgraph service deployment https://github.com/graphprotocol/contracts/blob/horizon/packages/horizon/tasks/deploy.ts#L93

Once we fix that we can get rid of this.

packages/horizon/scripts/e2e/pre-upgrade.ts
tmigone: ah cool. can we add a comment then explaining we need old ABIs?

packages/horizon/test/integration/during-transition-period/service-provider.test.ts
Maikol: Some of this tests depend on using the legacy storage that is accessed on HorizonStaking by doing

```solidity
if (_verifier == SUBGRAPH_DATA_SERVICE_ADDRESS) {
    ...
} else {
    ...
}
```

We probably have to move these tests to the subgraph service. 

Again because of the issue with the migrate script in horizon requiring the subgraph service the variable `SUBGRAPH_DATA_SERVICE_ADDRESS` in HorizonStaking is `!= address(0)`.

packages/horizon/test/integration/during-transition-period/slasher.test.ts
Maikol: Yes, it's legacy slashing.

packages/horizon/test/integration/during-transition-period/delegator.test.ts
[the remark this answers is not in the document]
Maikol: There's no changes to delegation during transition period or after transition period, that's why I didn't include most of this cases. We can definitely still do them if you think it's valuable. I like the idea of doing undelegate during the transition period and withdraw after.

packages/horizon/test/integration/during-transition-period/service-provider.test.ts
```diff
@@ -0,0 +1,376 @@
...
+      // Withdraw all thawed tokens
+      await withdraw({ horizonStaking, serviceProvider })
+
+      // Verify all tokens are withdrawn and transferred back to service provider
+      const balanceAfter = await graphToken.balanceOf(serviceProvider.address)
+      expect(balanceAfter).to.equal(balanceBefore + request1 + request2, 'Tokens were not transferred back to service provider')
+    })
+  })
+
+  describe('Existing Protocol Users', () => {
```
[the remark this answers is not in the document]
Maikol: > unstakes before transition period then withdraws during transition period

I believe this one is already implemented: `it(should allow service provider to withdraw their locked tokens after thawing period passes)` or am I missing something?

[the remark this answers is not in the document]
tmigone: looks like i missed that, oops!

packages/horizon/hardhat.config.ts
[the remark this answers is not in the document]
Maikol: It was just a mention that this `if` was for importing the SDK

packages/horizon/test/integration/during-transition-period/delegator.test.ts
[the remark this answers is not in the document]
tmigone: hmm good point! My completionist brain just wanted all things tested everywhere, but I agree there is no point if nothing changes. Then I think we just need to add this one?

- Delegate before transition period (existing user)
	- Undelegate during or after transition period and withdraw during or after transition period

PR #1112, opened 2025-02-17
Add prettier to `test` files

matiasedgeandnode:
I didn't add `test` to `solhint` because that lit up with +300 warnings. Probably best left for another PR.


matiasedgeandnode:
Should we do this for all `lint:sol` scripts?

tmigone (approved):
Looks good, i think we can address other changes as needed.

> lit up with +300 warnings

yikes

PR #1114, opened 2025-02-21
fix: only allow subgraph service as verifier during transition period

tmigone:


Maikol (approved):

packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -720,6 +720,11 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
         uint64 _thawingPeriod
     ) private {
         require(_tokens > 0, HorizonStakingInvalidZeroTokens());
+        // TODO: Remove this after the transition period - it prevents an early escape hatch for legacy allocations
+        require(
+            __DEPRECATED_thawingPeriod == 0 || _verifier == SUBGRAPH_DATA_SERVICE_ADDRESS,
```
pcarranzav, line 725: nit: slightly more gas efficient to check _verifier first, and only check the thawing period from storage otherwise

tmigone: fixed!

PR #1133, opened 2025-03-25
fix: remove redundant code (OZ N-01)

tmigone:


Maikol (approved):

Maikol:
I think this is fine but we'll reintroduce `DelegationWithBeneficiary` at some point and will have to add these back.

PR #1134, opened 2025-03-26
feat: implement Indexing Agreements [Pending Audit]

matiasedgeandnode:
Additional reading: [Indexing Payments technical specification v0.1](https://www.notion.so/edgeandnode/Indexing-Payments-technical-specification-v0-1-1c28686fc7c28026a07fde2458915d8f?pvs=4)


packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    function get(Manager storage self, bytes16 agreementId) external view returns (AgreementWrapper memory) {
+        AgreementWrapper memory wrapper = _get(self, agreementId);
+        require(wrapper.collectorAgreement.dataService == address(this), IndexingAgreementNotActive(agreementId));
+
+        return wrapper;
+    }
+
+    function _getManager() internal pure returns (Manager storage manager) {
```
pcarranzav, line 382: This is a good idea but I recommend using ERC7201 like the OZ contracts: https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/utils/PausableUpgradeable.sol#L19-L31

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -111,6 +111,14 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
         _;
     }
 
+    modifier onlyAuthorizedForProvisionHack(address caller, address serviceProvider) {
```
pcarranzav, line 114: Maybe just modify the original `onlyAuthorizedForProvision` to specify the caller?

packages/horizon/contracts/interfaces/IRecurringCollector.sol
```diff
@@ -0,0 +1,347 @@
...
+
+    /// @notice A representation of a signed Recurring Collection Agreement Upgrade (RCAU)
+    struct SignedRCAU {
+        // The RCAU
+        RecurringCollectionAgreementUpgrade rcau;
+        // Signature - 65 bytes: r (32 Bytes) || s (32 Bytes) || v (1 Byte)
+        bytes signature;
+    }
+
+    struct RecurringCollectionAgreementUpgrade {
```
pcarranzav, line 72: nit: I think `Update` is more correct than `Upgrade` for this (upgrade makes me think something improved, vs just a newer version of the terms)

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,472 @@
...
+            RecurringCollectorAgreementInvalidParameters("zero address")
+        );
+
+        // Agreement needs to end in the future
+        require(
+            _agreement.endsAt > block.timestamp,
+            RecurringCollectorAgreementInvalidParameters("endsAt not in future")
+        );
+
+        // Collection window needs to be at least 2 hours
```
pcarranzav, line 318: Any reason to enforce this value? I suspect some data services could allow much faster collection windows

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,472 @@
...
+     */
+    function _recoverRCAUSigner(SignedRCAU memory _signedRCAU) private view returns (address) {
+        bytes32 messageHash = _encodeRCAU(_signedRCAU.rcau);
+        return ECDSA.recover(messageHash, _signedRCAU.signature);
+    }
+
+    /**
+     * @notice See {IRecurringCollector.encodeRCA}
+     */
+    function _encodeRCA(RecurringCollectionAgreement memory _rca) private view returns (bytes32) {
```
pcarranzav, line 386: I'd suggest renaming to  `_encodeRCAHash` and same for RCAU, to make it clear this produces the hash that will be signed and not an abi-encoded RCA

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,472 @@
...
+
+    /**
+     * @notice Requires that the collection params are valid.
+     */
+    function _requireValidCollect(
+        AgreementData memory _agreement,
+        bytes16 _agreementId,
+        uint256 _tokens
+    ) private view returns (uint256) {
+        uint256 collectionSeconds = _agreement.state == AgreementState.CanceledByPayer
```
pcarranzav, line 340: It's not super clear to me if this will somehow ensure in the CanceledByPayer case only one last collection is allowed?

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -129,6 +131,19 @@ contract DisputeManager is
         return _createIndexingDisputeWithAllocation(msg.sender, disputeDeposit, allocationId, poi);
     }
 
+    /// @inheritdoc IDisputeManager
+    function createIndexingFeeDisputeV1(
+        bytes16 agreementId,
+        bytes32 poi,
```
pcarranzav, line 137: should we also include the block at which the poi was posted? that way multiple collections with the same poi/entities can be disputed separately

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
matiasedgeandnode: I was actually thinking of moving this into a func in a new lib, and have the modifiers call the lib func. Didn't want to change the original modifier, because it's called 6 times in the service, and it seems safer as it is, without specifying the caller. Thoughts?

packages/horizon/contracts/interfaces/IRecurringCollector.sol
matiasedgeandnode: happy to change this.

packages/horizon/contracts/utilities/GraphDirectory.sol
```diff
@@ -131,6 +131,14 @@ abstract contract GraphDirectory {
         );
     }
 
+    /**
+     * @notice Get the Epoch Manager contract
+     * @return The Epoch Manager contract
+     */
+    function graphEpochManager() external view returns (IEpochManager) {
```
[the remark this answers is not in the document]
matiasedgeandnode: Well, because this is required by `IndexingAgreement.collect( ... )` which is a library. And so I believe the options are to inject the dependency, or to expose the `graphEpochManager()` externally. Thoughts?

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: This limits the collection window for canceled agreements by payers. i.e. if the agreement has been canceled by the payer, the service provider is allowed to collect for work done up to the cancelation time. I agree, it isn't super clear. Maybe a comment would help?

packages/subgraph-service/contracts/DisputeManager.sol
matiasedgeandnode: Absolutely, great catch, thanks.

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.27;
+
+import { IndexingAgreement } from "./IndexingAgreement.sol";
+
+library UnsafeDecoder {
```
pcarranzav, line 6: Why is it unsafe?

pcarranzav: And I'm not sure I see the benefit of implementing these separately from Decoder.sol, is there a size/performance improvement?

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,459 @@
...
+        bytes16 _agreementId,
+        IRecurringCollector.AgreementData memory _agreement,
+        uint256 _entities
+    ) private view returns (uint256) {
+        IndexingAgreementTermsV1 memory termsV1 = _manager.termsV1[_agreementId];
+
+        uint256 collectionSeconds = block.timestamp;
+        collectionSeconds -= _agreement.lastCollectionAt > 0 ? _agreement.lastCollectionAt : _agreement.acceptedAt;
+
+        // FIX-ME: this is bad because it encourages indexers to collect at max seconds allowed to maximize collection.
```
pcarranzav, line 429: I think it's reasonable imo, they are allowed to do it and in the long term they would still collect the same amount, won't they? Do you see an alternative approach?

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
pcarranzav: I see... I'd be okay with both options, I don't think it's too bad to modify the original, but moving to a new lib sounds good too

packages/horizon/contracts/utilities/GraphDirectory.sol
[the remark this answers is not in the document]
pcarranzav: ah, I see... I personally would find dependency injection nicer, but I'm okay with whichever you prefer

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
pcarranzav: hmm it makes sense but I would prefer setting this at the data service level rather than in the collector, as different data services might have different timing requirements

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
pcarranzav: Ooh I see, so it would underflow if called for a second time I think? Maybe explicit handling of canceledAt < lastCollectionAt would be a bit cleaner?
But yeah in any case, a comment would be useful 👍

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: hmm... lets chat.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: How about `_hashRCA` or even `_hash712RCA` / `_hashRCA712`?

pcarranzav: `_hashRCA` sgtm 👍

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
matiasedgeandnode: naming could use some improvement. It's "unsafe" in the sense that it reverts with a generic error.

matiasedgeandnode: `Decoder` and `UnsafeDecoder` are two separate libs, because:
* `Decoder` is a lib to reduce size in the service.
* `try / catch` can only be used on external function call.
* A lib can't call itself externally.
All of these together resulted in the two separate implementations. Maybe there is a better way to do it that I'm missing?

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: I'll remove the comment. I think in the long run they'll probably collect roughly the same amount.

packages/horizon/contracts/utilities/GraphDirectory.sol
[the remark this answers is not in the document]
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,472 @@
...
+        address signer = _recoverRCAUSigner(_signedRCAU);
+        require(_isAuthorized(_payer, signer), RecurringCollectorInvalidSigner());
+
+        return signer;
+    }
+
+    /**
+     * @notice Gets an agreement to be updated.
+     */
+    function _getForUpdateAgreement(bytes16 _agreementId) private view returns (AgreementData storage) {
```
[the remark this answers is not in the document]
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,472 @@
...
+        require(
+            agreement.state == AgreementState.Accepted,
+            RecurringCollectorAgreementIncorrectState(agreementId, agreement.state)
+        );
+        require(
+            agreement.dataService == msg.sender,
+            RecurringCollectorDataServiceNotAuthorized(agreementId, msg.sender)
+        );
+        agreement.canceledAt = uint64(block.timestamp);
+        agreement.state = by == CancelAgreementBy.Payer
```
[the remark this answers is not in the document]
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: Added `MIN_SECONDS_COLLECTION_WINDOW` and set it to `600`

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: fixed

matiasedgeandnode:
@pcarranzav ready for another round. Besides fixes to your comments, I've made the subgraph-service contract size < 24kb. Probably worth discussing the approach taken. I've left the original implementation commented out (to be removed later) so that it's obvious what I've changed. Also note that the code has been moved into libs only till it fits and there are more gains to be had if desired.

packages/subgraph-service/contracts/DisputeManager.sol
matiasedgeandnode: done

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.27;
+
+import { ProvisionTracker } from "./ProvisionTracker.sol";
+import { IDataServiceFees } from "../interfaces/IDataServiceFees.sol";
+import { IHorizonStaking } from "../../interfaces/IHorizonStaking.sol";
+import { LinkedList } from "../../libraries/LinkedList.sol";
+
+library DataServiceFeesLib {
```
tmigone, line 9:
```suggestion
library StakeClaims {
```

How about we rename this? I'm personally not very fond of the `Lib` naming approach, and I feel we can be more descriptive with the name.

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
```diff
@@ -0,0 +1,111 @@
...
+        return (false, _acc);
+    }
+
+    /**
+     * @notice Builds a stake claim ID
+     * @param serviceProvider The address of the service provider
+     * @param nonce A nonce of the stake claim
+     * @return The stake claim ID
+     */
+    function _buildStakeClaimId(address serviceProvider, uint256 nonce) internal view returns (bytes32) {
```
tmigone, line 108: Since this is now a linked lib I think it could be useful to expose this internal function with an external getter as well.

packages/horizon/contracts/data-service/libraries/ProvisionManagerLib.sol
tmigone, line 1: This one is so small that it's a shame to make it a lib. If we move the rest of the stake claim functionality to the lib we could probably add this back to `ProvisionManager`?

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity 0.8.27;
+
+import { ProvisionTracker } from "./ProvisionTracker.sol";
+import { IDataServiceFees } from "../interfaces/IDataServiceFees.sol";
```
tmigone, line 5: I would remove the dependency with `IDataServiceFees` any event or error that we use from there I think should probably instead live in this lib contract. Note that if we move any errors into this file we would need to rename their prefix `IDataServiceFees.DataServiceFeesClaimNotFound` -> `StakeClaimsClaimNotFound`

packages/horizon/contracts/data-service/libraries/ProvisionTracker.sol
tmigone, line 1: As far as i understand in this file there is 1 external fn but 2 internal functions. I think we should either:
- make all functions external, then use it as a linked lib
- make all functions internal, then use it embedded

I think this lib is no longer being imported from a contract with size problems so we could do either and it should be fine?

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -124,10 +122,14 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
...
+        this.requireValidProvision(serviceProvider);
+        _;
+    }
+
+    function requireValidProvision(address serviceProvider) external view {
         IHorizonStaking.Provision memory provision = _getProvision(serviceProvider);
         _checkProvisionTokens(provision);
         _checkProvisionParameters(provision, false);
-        _;
     }
```
tmigone, line 133: Why do we need this external function?

Also the usage of `this.` forces it to make an external call to itself, so a public function would be better here it would remove that external call. (still think we probably don't need this as an externally callable fn)

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
```diff
@@ -320,4 +320,8 @@ abstract contract ProvisionManager is Initializable, GraphDirectory, ProvisionMa
     function _checkValueInRange(uint256 _value, uint256 _min, uint256 _max, bytes memory _revertMessage) private pure {
         require(_value.isInRange(_min, _max), ProvisionManagerInvalidValue(_revertMessage, _value, _min, _max));
     }
+
+    function _requireLTE(uint256 _min, uint256 _max) private pure {
```
tmigone, line 324: Missing natspec

packages/horizon/contracts/libraries/LinkedList.sol
tmigone, line 1: Same comment as the previous lib, I think we should have all internal or all external fns. Having a mix looks hacky (though it works fine). 
I think this lib has the extra challenge with the fns as parameters, so it might be complex to make all external. If we have the space (which i think we do have now) I would suggest we keep this lib as all internal fns and use it as an embedded library.

tmigone (commented):
Only reviewed changes to `horizon` package (not including DIPs stuff).

packages/horizon/contracts/data-service/libraries/ProvisionTracker.sol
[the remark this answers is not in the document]
matiasedgeandnode: hmmm... when I first coded this, having all external increased the contract size. But now it shrinked. Done.

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
```diff
@@ -0,0 +1,111 @@
...
+     * - The associated provision must have enough available tokens to lock the stake.
+     *
+     * Emits a {StakeClaimLocked} event.
+     *
+     * @param _serviceProvider The address of the service provider
+     * @param _tokens The amount of tokens to lock in the claim
+     * @param _unlockTimestamp The timestamp when the tokens can be released
+     */
+    function lockStake(
+        uint32 _delegationRatio,
```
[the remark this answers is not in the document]
matiasedgeandnode: indeed, fixed. Thanks

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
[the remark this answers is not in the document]
matiasedgeandnode: Happy to do that, initially stopped as soon as the contract was small enough, but if we are happy with this approach, I'd be glad to go all-in.

packages/subgraph-service/contracts/libraries/AllocationManagerLib.sol
```diff
@@ -0,0 +1,295 @@
...
+import { IHorizonStaking } from "@graphprotocol/horizon/contracts/interfaces/IHorizonStaking.sol";
+import { IGraphPayments } from "@graphprotocol/horizon/contracts/interfaces/IGraphPayments.sol";
+import { ProvisionTracker } from "@graphprotocol/horizon/contracts/data-service/libraries/ProvisionTracker.sol";
+import { PPMMath } from "@graphprotocol/horizon/contracts/libraries/PPMMath.sol";
+
+import { Allocation } from "../libraries/Allocation.sol";
+import { LegacyAllocation } from "../libraries/LegacyAllocation.sol";
+import { AllocationManager } from "../utilities/AllocationManager.sol";
+
+library AllocationManagerLib {
```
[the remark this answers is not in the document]
tmigone: How about `AllocationHandler.sol`?

packages/subgraph-service/contracts/libraries/AllocationManagerLib.sol
```diff
@@ -0,0 +1,295 @@
...
+        address _allocationId,
+        bytes memory _proof
+    ) private pure {
+        address signer = ECDSA.recover(_encodeAllocationProof, _proof);
+        require(
+            signer == _allocationId,
+            AllocationManager.AllocationManagerInvalidAllocationProof(signer, _allocationId)
+        );
+    }
+}
```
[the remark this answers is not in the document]
tmigone: ok i see now why 😄

packages/horizon/contracts/data-service/libraries/ProvisionTracker.sol
matiasedgeandnode: It is imported (indirectly) by the SubgraphService. As mentioned outside of this PR, we found enough space elsewhere that we can revert the changes to this lib if desired.

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
matiasedgeandnode: The extension needs to call this. It does so by externally calling the base (itself) which inherits via `SubgraphService -> DataService -> ProvisionManager`.

matiasedgeandnode: In any case, I've removed the `this.` call. Adding the `internal` function shrunk the contract by another 0.36kb.

packages/subgraph-service/contracts/SubgraphServiceExtension.sol
```diff
@@ -0,0 +1,84 @@
...
+    modifier onlyValid(address indexer) {
+        ProvisionManagerLib.requireAuthorizedForProvision(
+            IHorizonStaking(_getBase().getGraphStaking()),
+            indexer,
+            address(this),
+            msg.sender
+        );
+        _getBase().requireValidProvision(indexer);
+        _getBase().requireRegisteredIndexer(indexer);
+        _;
```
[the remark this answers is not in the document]
matiasedgeandnode: The issue I believe comes up with that strategy is that if you deploy horizon as is, then I can't release this change, because the storage slots won't map the same way, right?

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -45,13 +49,14 @@ contract SubgraphService is
...
     using TokenUtils for IGraphToken;
+    using IndexingAgreement for IndexingAgreement.Manager;
 
     /**
      * @notice Checks that an indexer is registered
      * @param indexer The address of the indexer
      */
     modifier onlyRegisteredIndexer(address indexer) {
-        require(indexers[indexer].registeredAt != 0, SubgraphServiceIndexerNotRegistered(indexer));
+        this.requireRegisteredIndexer(indexer);
```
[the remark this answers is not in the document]
matiasedgeandnode: Indeed. Fixed. Thanks!

packages/subgraph-service/contracts/libraries/AllocationManagerLib.sol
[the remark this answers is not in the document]
matiasedgeandnode: I think it was an oversight. Removed it and it works just fine. Hmmm... I'm confused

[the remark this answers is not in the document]
matiasedgeandnode: Sorry, got confused, I removed the one from `AllocationManager`, kept this one. We are good.

packages/subgraph-service/contracts/utilities/AllocationManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: Done, thanks.

packages/horizon/contracts/interfaces/IRecurringCollector.sol
matiasedgeandnode: done

packages/subgraph-service/contracts/libraries/AllocationManagerLib.sol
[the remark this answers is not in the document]
matiasedgeandnode: I believe it can. We can either expose `_hashTypedDataV4` externally or we can manage the domain separator in the library a use `MessageHashUtils.toTypedDataHash`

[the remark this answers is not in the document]
tmigone: I think the problem is that the library cannot inherit the EIP712 OZ library.

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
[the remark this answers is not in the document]
tmigone: cant we move `_releaseStake()`, `_deleteStakeClaim()` and `_getNextStakeClaim()`?

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
tmigone: ah yes that makes sense. to avoid name collision we can rename the struct to `State` similar to what we have with `Allocation.State` and a few others.

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
matiasedgeandnode: Agreed, it was meant to be a placeholder until we agreed on the validity of the approach.

packages/subgraph-service/contracts/libraries/SubgraphServiceLib.sol
tmigone, line 1: Can we put this back in the SubgraphService given we now have space?

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
tmigone: im confused by this as well, cant we put everything into a Decoder library that has only external functions?

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -226,13 +235,14 @@ contract SubgraphService is
             _allocations.get(allocationId).indexer == indexer,
             SubgraphServiceAllocationNotAuthorized(indexer, allocationId)
         );
+        _cancelAllocationIndexingAgreement(allocationId);
```
tmigone, line 238: Can we explicitly check first if an agreement exists before attempting to cancel it? 
I've walked down the code and I think it's just a no-op if there is no agreement, but there are several contracts involved all which can change in the future, so better be explicit and safe.

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -302,6 +319,7 @@ contract SubgraphService is
         Allocation.State memory allocation = _allocations.get(allocationId);
         require(allocation.isStale(maxPOIStaleness), SubgraphServiceCannotForceCloseAllocation(allocationId));
         require(!allocation.isAltruistic(), SubgraphServiceAllocationIsAltruistic(allocationId));
+        _cancelAllocationIndexingAgreement(allocationId);
```
tmigone, line 322: Same as before, can we explicitly check if an agreement exists?

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+        uint256 currentEpoch;
+        bytes data;
+    }
+
+    /// @custom:storage-location erc7201:graphprotocol.subgraph-service.storage.Manager.IndexingAgreement
+    struct Manager {
+        mapping(bytes16 => State) agreements;
+        mapping(bytes16 agreementId => IndexingAgreementTermsV1 data) termsV1;
+        mapping(address allocationId => bytes16 agreementId) allocationToActiveAgreementId;
+    }
```
tmigone, line 82: It's not immediately obvious what this `Manager` is for, I would consider renaming to `StorageManager` and `_getStorageManager()`

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+     * @notice Thrown when trying to interact with an agreement with an invalid version
+     * @param version The invalid version
+     */
+    error InvalidIndexingAgreementVersion(IndexingAgreementVersion version);
+
+    /**
+     * @notice Thrown when an agreement is not for the subgraph data service
+     * @param wrongDataService The wrong data service
+     */
+    error IndexingAgreementWrongDataService(address wrongDataService);
```
tmigone, line 174: Might be dumb because it's trivial to get it, but i would add the emitter data service (`address(this)`) to this error. It can be helpful during debugging. Also make both indexed

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
tmigone, line 1: There are several natspec omissions in the contract, be sure to check those.

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+                allocationId,
+                allocation.subgraphDeploymentId
+            )
+        );
+
+        require(
+            self.allocationToActiveAgreementId[allocationId] == bytes16(0),
+            AllocationAlreadyHasIndexingAgreement(allocationId)
+        );
+        self.allocationToActiveAgreementId[allocationId] = signedRCA.rca.agreementId;
```
tmigone, line 255: This ensures only one agreement per allo right? I'd add a comment

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+     * @notice Thrown when the agreement is already accepted
+     * @param agreementId The agreement ID
+     */
+    error IndexingAgreementAlreadyAccepted(bytes16 agreementId);
+
+    /**
+     * @notice Thrown when an allocation already has an active agreement
+     * @param allocationId The allocation ID
+     */
+    error AllocationAlreadyHasIndexingAgreement(address allocationId);
```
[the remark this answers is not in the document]
tmigone: that is one hell of an error name btw 😃

tmigone, line 198:
```suggestion
    error IndexingAgreementAllocationAlreadyHasIndexingAgreement(address allocationId);
```

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+ * @title RecurringCollector contract
+ * @dev Implements the {IRecurringCollector} interface.
+ * @notice A payments collector contract that can be used to collect payments using a RCA (Recurring Collection Agreement).
+ * @custom:security-contact Please email security+contracts@thegraph.com if you find any
+ * bugs. We may have an active bug bounty program.
+ */
+contract RecurringCollector is EIP712, GraphDirectory, Authorizable, IRecurringCollector {
+    using PPMMath for uint256;
+
+    uint32 public constant MIN_SECONDS_COLLECTION_WINDOW = 600;
```
tmigone, line 24: natspec

packages/horizon/contracts/interfaces/IRecurringCollector.sol
```diff
@@ -0,0 +1,366 @@
...
+     * @param agreementId The agreement ID
+     * @param unauthorizedDataService The address of the unauthorized data service
+     */
+    error RecurringCollectorDataServiceNotAuthorized(bytes16 agreementId, address unauthorizedDataService);
+
+    /**
+     * Thrown when interacting with an agreement with an elapsed deadline
+     * @param deadline The elapsed deadline timestamp
+     */
+    error RecurringCollectorAgreementDeadlineElapsed(uint64 deadline);
```
tmigone, line 243: nit: another silly one, consider adding the current timestamp as well here. Again, useful for debugging.

packages/horizon/contracts/interfaces/IRecurringCollector.sol
```diff
@@ -0,0 +1,366 @@
...
+     * @param agreementId The agreement ID
+     * @param incorrectState The incorrect state
+     */
+    error RecurringCollectorAgreementIncorrectState(bytes16 agreementId, AgreementState incorrectState);
+
+    /**
+     * Thrown when accepting or upgrading an agreement with invalid parameters
+     * @param message A descriptive error message
+     */
+    error RecurringCollectorAgreementInvalidParameters(string message);
```
tmigone, line 288: This pattern has proven to be an inconvenience for subgraph development, also the message string is not always parsed as text by block explorers and other tools which is annoying. 

It's a pain but it's better to have dedicated errors for each case.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+        agreement.state = AgreementState.Accepted;
+        agreement.dataService = signedRCA.rca.dataService;
+        agreement.payer = signedRCA.rca.payer;
+        agreement.serviceProvider = signedRCA.rca.serviceProvider;
+        agreement.endsAt = signedRCA.rca.endsAt;
+        agreement.maxInitialTokens = signedRCA.rca.maxInitialTokens;
+        agreement.maxOngoingTokensPerSecond = signedRCA.rca.maxOngoingTokensPerSecond;
+        agreement.minSecondsPerCollection = signedRCA.rca.minSecondsPerCollection;
+        agreement.maxSecondsPerCollection = signedRCA.rca.maxSecondsPerCollection;
+        _requireValidAgreement(agreement);
```
tmigone, line 109: nit: usually you'd check the inputs before doing any storage ops, consider changing the order of operations here.

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+
+        return collectionSeconds * (termsV1.tokensPerSecond + termsV1.tokensPerEntityPerSecond * _entities);
+    }
+
+    function _isActive(AgreementWrapper memory wrapper) private view returns (bool) {
+        return
+            wrapper.collectorAgreement.dataService == address(this) &&
+            wrapper.collectorAgreement.state == IRecurringCollector.AgreementState.Accepted &&
+            wrapper.agreement.allocationId != address(0);
+    }
```
tmigone, line 498: Took me a bit to figure out why we need all this conditions, worth documenting this. Namely that for an agreement to be considered valid it needs to:
- be registered and accepted in the collector for this specific data service
- be registered in this contract

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+            agreement.maxSecondsPerCollection
+        );
+    }
+
+    /**
+     * @notice Cancel an indexing agreement.
+     * See {IRecurringCollector.cancel}.
+     * @dev Caller must be the data service for the agreement.
+     */
+    function cancel(bytes16 agreementId, CancelAgreementBy by) external {
```
tmigone, line 130: Wasted opportunity to utilize the term `canceloor`

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+
+    function _cancel(
+        Manager storage _manager,
+        bytes16 _agreementId,
+        State memory _agreement,
+        IRecurringCollector.AgreementData memory _collectorAgreement,
+        IRecurringCollector.CancelAgreementBy _cancelBy
+    ) private {
+        delete _manager.allocationToActiveAgreementId[_agreement.allocationId];
```
tmigone, line 465: Should we delete the agreement as well? In `accept()` we are writing to storage two times one for `allocationToActiveAgreementId` and the other for `agreements`. Seems like we should delete both or none?

A side effect of not deleting the agreement is that the portion of the active check in `_isActive()` that is relevant to this contract will still have the agreement as valid. This doesn't matter though because the rest of the checks in there will make sure it's overall invalid. But still wanted to point it out.

packages/horizon/contracts/interfaces/IRecurringCollector.sol
```diff
@@ -0,0 +1,366 @@
...
+    // @notice The state of an agreement
+    enum AgreementState {
+        NotAccepted,
+        Accepted,
+        CanceledByServiceProvider,
+        CanceledByPayer
+    }
+
+    // @notice The party that can cancel an agreement
+    enum CancelAgreementBy {
```
tmigone, line 25: We probably want to add another party here to represent agreements that got canceled by a third party when force closing a stale allocation.

packages/subgraph-service/contracts/SubgraphService.sol
tmigone, line 322: Another thing, ideally we want to differentiate this cancellation (initiated by a third party) from regular cancellation due to allocation closing (initiated by the service provider).
Currently they are both using `IRecurringCollector.CancelAgreementBy.ServiceProvider`

packages/horizon/contracts/interfaces/IRecurringCollector.sol
pcarranzav: hmm interesting edge case... though I would consider that as cancelled by the indexer?

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+            RecurringCollectorAgreementIncorrectState(_params.agreementId, agreement.state)
+        );
+
+        require(
+            msg.sender == agreement.dataService,
+            RecurringCollectorDataServiceNotAuthorized(_params.agreementId, msg.sender)
+        );
+
+        require(
+            agreement.endsAt >= block.timestamp,
```
pcarranzav, line 267: This blocks collections after the agreement ended, but I wonder if it should still be valid to collect after that? In which case we could set the collectionEnd to endsAt in _requireValidCollect?

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -450,6 +465,85 @@ contract DisputeManager is
...
+
+        // Only one dispute at a time
+        require(!isDisputeCreated(disputeId), DisputeManagerDisputeAlreadyCreated(disputeId));
+
+        // The indexer must be disputable
+        IHorizonStaking.Provision memory provision = _graphStaking().getProvision(
+            wrapper.collectorAgreement.serviceProvider,
+            address(_getSubgraphService())
+        );
+        require(provision.tokens != 0, DisputeManagerZeroTokens());
```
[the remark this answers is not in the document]
pcarranzav: I think so, @tmigone wdyt?

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
pcarranzav: I would be fine with just using the "unsafe" variant and using the native errors from abi.decode

packages/subgraph-service/contracts/SubgraphService.sol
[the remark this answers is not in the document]
tmigone: no longer relevant as we dont have an extension anymore

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
tmigone: I think the benefit of catching and re throwing is that we can be more descriptive with the error. iirc abi.decode tends to fail with `Reverted without reason`. I don't mind either approach but I think if we try/catch we could combine both libraries into a single one.

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -256,6 +267,9 @@ contract SubgraphService is
      *      - address `allocationId`: The id of the allocation
      *      - bytes32 `poi`: The POI being presented
      *      - bytes `poiMetadata`: The metadata associated with the POI. See {AllocationManager-_collectIndexingRewards} for more details.
+     *    - For indexing fees:
+     *      - bytes16 `agreementId`: The id of the indexing agreement
+     *      - bytes `agreementCollectionMetadata`: The metadata required by the indexing agreement version.
      */
     /// @inheritdoc IDataService
     function collect(
```
tmigone, line 275: Can we add some comment for this function's natspec explaining the stake lock mechanism? 

```
Note that collecting any type of payment will require locking provisioned stake as collateral for a period of time. All types of payment share the same pool of provisioned stake however they each have separate accounting:
- Indexing rewards can make full use of the available stake
- Query and indexing fees share the pool, combined they can also make full use of the available stake
```

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+            InvalidIndexingAgreementVersion(wrapper.agreement.version)
+        );
+
+        (uint256 entities, bytes32 poi, uint256 poiEpoch) = Decoder.decodeCollectIndexingFeeDataV1(params.data);
+
+        uint256 expectedTokens = (entities == 0 && poi == bytes32(0))
+            ? 0
+            : _tokensToCollect(self, params.agreementId, wrapper.collectorAgreement, entities);
+
+        uint256 tokensCollected = _directory().recurringCollector().collect(
```
tmigone, line 410: Is there any difference between `expectedTokens` and `tokensCollected`? Since there is no validation here and the variable names are suggestive I would add a comment to clarify what the expectation is.

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+        Manager storage self,
+        mapping(address allocationId => Allocation.State allocation) storage allocations,
+        CollectParams memory params
+    ) external returns (address, uint256) {
+        AgreementWrapper memory wrapper = _get(self, params.agreementId);
+        Allocation.State memory allocation = allocations.requireValidAllocation(
+            wrapper.agreement.allocationId,
+            wrapper.collectorAgreement.serviceProvider
+        );
+        require(_isActive(wrapper), IndexingAgreementNotActive(params.agreementId));
```
tmigone, line 397: This implies that indexers won't be able to collect on a closed allocation (as agreements are automatically cancelled when the associated allocation is closed). Is this intentional? If it is, then we should document it properly as I'd say that is not the "expected behavior" if you consider how query fees currently work.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+     * @notice Initiate a payment collection through the payments protocol.
+     * See {IGraphPayments.collect}.
+     * @dev Caller must be the data service the RCA was issued to.
+     */
+    function collect(IGraphPayments.PaymentTypes paymentType, bytes calldata data) external returns (uint256) {
+        require(
+            paymentType == IGraphPayments.PaymentTypes.IndexingFee,
+            RecurringCollectorInvalidPaymentType(paymentType)
+        );
+        try this.decodeCollectData(data) returns (CollectParams memory collectParams) {
```
tmigone, line 65:
```suggestion
        try decodeCollectData(data) returns (CollectParams memory collectParams) {
```

packages/subgraph-service/contracts/libraries/UnsafeDecoder.sol
matiasedgeandnode: I believe we can't combine both libraries into one. As I mentioned before `try / catch` needs to call an external function and a library can't call it's own external functions. I'd prefer to keep the custom errors because the amount of time I spent debugging `Reverted without reason` dwarfs the added complexity of having two libs. Would be very happy to find a better name for either lib though.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+        address controller,
+        uint256 revokeSignerThawingPeriod
+    ) EIP712(eip712Name, eip712Version) GraphDirectory(controller) Authorizable(revokeSignerThawingPeriod) {}
+
+    /**
+     * @notice Initiate a payment collection through the payments protocol.
+     * See {IGraphPayments.collect}.
+     * @dev Caller must be the data service the RCA was issued to.
+     */
+    function collect(IGraphPayments.PaymentTypes paymentType, bytes calldata data) external returns (uint256) {
```
tmigone, line 60: Can we document what the requirements are for a successful collection? Mostly interested in allocation and agreement status.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+            tokensToCollect = _requireValidCollect(agreement, _params.agreementId, _params.tokens);
+
+            _graphPaymentsEscrow().collect(
+                IGraphPayments.PaymentTypes.IndexingFee,
+                agreement.payer,
+                agreement.serviceProvider,
+                tokensToCollect,
+                agreement.dataService,
+                _params.dataServiceCut,
+                agreement.serviceProvider // FIX-ME
```
tmigone, line 282: 👀

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -503,6 +518,86 @@ contract DisputeManager is
...
+            abi.encodePacked(
+                "IndexingFeeDisputeWithAgreement",
+                _agreementId,
+                wrapper.collectorAgreement.serviceProvider,
+                wrapper.collectorAgreement.payer,
+                _poi,
+                _entities,
+                _blockNumber
+            )
+        );
```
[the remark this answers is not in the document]
matiasedgeandnode: hmmm... this got me thinking. `serviceProvider` and `payer` do seem to be redundant. So does `entities`. But also `poi`? All I really need to identify the disputed collection is the `agreementId` and the `blockNumber`?

packages/subgraph-service/contracts/libraries/SubgraphServiceLib.sol
matiasedgeandnode: I can move it into `IndexingAgreement` library, which is where it's used, but it's not super elegant. This fn was never part of `SubgraphService`. If I move it into the `SubgraphService` I'd need to `external` it, so that `IndexingAgreement` can call it.

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: To check without trusting the library, I'd have to duplicate `IndexingAgreement` specific logic in the `SubgraphService` contract. Specifically:
```
function _getManager() internal pure returns (Manager storage $) {
    // solhint-disable-next-line no-inline-assembly
    assembly {
        $.slot := INDEXING_AGREEMENT_MANAGER_STORAGE_LOCATION
    }
}
```
and 
```
function _get(Manager storage self, bytes16 agreementId) private view returns (AgreementWrapper memory) {
    return
        AgreementWrapper({
            agreement: self.agreements[agreementId],
            collectorAgreement: _directory().recurringCollector().getAgreement(agreementId)
        });
}
```
and
```
function _isActive(AgreementWrapper memory wrapper) private view returns (bool) {
    return
        wrapper.collectorAgreement.dataService == address(this) &&
        wrapper.collectorAgreement.state == IRecurringCollector.AgreementState.Accepted &&
        wrapper.agreement.allocationId != address(0);
}
```
plus the actual check. I don't think this would be much better. Maybe we can discuss off-PR?

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+     * @notice Thrown when an agreement and the allocation correspond to different deployment IDs
+     * @param agreementDeploymentId The agreement's deployment ID
+     * @param allocationId The allocation ID
+     * @param allocationDeploymentId The allocation's deployment ID
+     */
+    error IndexingAgreementDeploymentIdMismatch(
+        bytes32 agreementDeploymentId,
+        address allocationId,
+        bytes32 allocationDeploymentId
+    );
```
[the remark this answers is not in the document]
matiasedgeandnode: hmmm... the agreement doesn't have an allocationId at this point...

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: I've added the comment. I'd rather avoid moving this around because I'd have to re-write a* test while saving gas on reverts seems too much, hopefully users are simulating txs?

packages/subgraph-service/contracts/DisputeManager.sol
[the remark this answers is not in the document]
tmigone: It sounds reasonable. I looked at previous audits to see if this was discussed but did not find anything.

Fixed it here: [`d89440a` (#1183)](https://github.com/graphprotocol/contracts/pull/1183/commits/d89440a95ffaf9a9abfc1a6836399e3189e66e50)

Going to include it in the fix review OZ is going to do for us.

packages/horizon/contracts/interfaces/IRecurringCollector.sol
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: plenty of time still

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+            wrapper.agreement,
+            wrapper.collectorAgreement,
+            IRecurringCollector.CancelAgreementBy.ServiceProvider
+        );
+    }
+
+    function cancelByPayer(Manager storage self, bytes16 agreementId) external {
+        AgreementWrapper memory wrapper = _get(self, agreementId);
+        require(_isActive(wrapper), IndexingAgreementNotActive(agreementId));
+        require(
```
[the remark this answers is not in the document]
matiasedgeandnode: `SubgraphService` is calling `IndexingAgreements` via `delegatecall`.

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: Yes. `expectedTokens` is what the data service says the max payout should be, based on it's concern of entities and time, without looking at the RCA terms. `tokensCollected` will be `LTE` `expectedTokens` because the collector will further narrow that down based on the max tokens per second params and other RCA terms.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode:
```
Compiler run failed:
Error (2536): Try can only be used with external function calls and contract creation calls.
  --> node_modules/@graphprotocol/horizon/contracts/payments/collectors/RecurringCollector.sol:65:13:
   |
65 |         try decodeCollectData(data) returns (CollectParams memory collectParams) {
   |             ^^^^^^^^^^^^^^^^^^^^^^^

Error: Compilation failed
```

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -370,6 +388,112 @@ contract SubgraphService is
...
+    function acceptIndexingAgreement(
+        address allocationId,
+        IRecurringCollector.SignedRCA calldata signedRCA
+    )
+        external
+        whenNotPaused
+        onlyAuthorizedForProvision(signedRCA.rca.serviceProvider)
+        onlyValidProvision(signedRCA.rca.serviceProvider)
+        onlyRegisteredIndexer(signedRCA.rca.serviceProvider)
+    {
```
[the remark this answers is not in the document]
matiasedgeandnode: I can see the benefit of that, though for me it's outweighed by the added complexity. i.e. The caller needs to pass in an extra param, needs to match what's in the RCA, and the contract need to check that that's actually the case.

`update` and `cancel` need the `indexer` to be passed in explicitly because it can't be inferred from other params.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
```diff
@@ -0,0 +1,481 @@
...
+            by
+        );
+    }
+
+    /**
+     * @notice Update an indexing agreement.
+     * See {IRecurringCollector.update}.
+     * @dev Caller must be the data service for the agreement.
+     */
+    function update(SignedRCAU calldata signedRCAU) external {
```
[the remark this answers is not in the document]
matiasedgeandnode: yes, this is expected.

packages/subgraph-service/contracts/DisputeManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: I'll port the fix...

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: done

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: done

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: I've added the note verbatim, but it doesn't ring true to me.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: Need to FIX-ME...

matiasedgeandnode:
Still missing a lot of natspec (and get clean(ish) `pnpm lint:sol:natspec`?)

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -425,6 +511,14 @@ contract SubgraphService is
         return _isOverAllocated(indexer, _delegationRatio);
     }
 
+    function getGraphStaking() external view returns (address) {
+        return address(_graphStaking());
+    }
```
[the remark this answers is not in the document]
tmigone: This function can now be deleted

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
tmigone: Sounds reasonable!

packages/subgraph-service/contracts/SubgraphService.sol
tmigone: you lock stake a priori by means of allocations, perhaps the wording is not the best here.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
tmigone: ahhh i hate try/catch

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: yeah ok, I'm happy as is.

packages/horizon/contracts/data-service/extensions/DataServiceFees.sol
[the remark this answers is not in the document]
matiasedgeandnode: I've done this now. Note that it added `0.150kb`.  Also, `_releaseStake` is still unchanged.

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
matiasedgeandnode: hmmm... I actually prefer it this way. It's smaller by 0.077kb and I prefer modifiers calling internal functions rather than inlined unless they are one liners.

packages/subgraph-service/contracts/SubgraphService.sol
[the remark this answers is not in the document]
matiasedgeandnode: done

packages/subgraph-service/contracts/DisputeManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: Yeah, you could present N ones, but I believe only one would collect fees. If we take that argument I'd need to include entities as well, because you could present the same poi twice with different entities.

packages/subgraph-service/contracts/DisputeManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: Let Trust know about this change.

packages/subgraph-service/contracts/DisputeManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: keep entities and POI

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: comment why this delete is here. i.e. to allow for allocation to be re-assigned

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
```diff
@@ -0,0 +1,519 @@
...
+            wrapper.collectorAgreement.serviceProvider
+        );
+        require(_isActive(wrapper), IndexingAgreementNotActive(params.agreementId));
+
+        require(
+            wrapper.agreement.version == IndexingAgreementVersion.V1,
+            InvalidIndexingAgreementVersion(wrapper.agreement.version)
+        );
+
+        (uint256 entities, bytes32 poi, uint256 poiEpoch) = Decoder.decodeCollectIndexingFeeDataV1(params.data);
```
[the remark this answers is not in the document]
matiasedgeandnode: add, don't replace.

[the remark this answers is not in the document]
matiasedgeandnode: actually shove it into metadata

packages/horizon/contracts/data-service/libraries/ProvisionManagerLib.sol
matiasedgeandnode: lib is gone.

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: see other thread.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: done, please double-check

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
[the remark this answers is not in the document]
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: done

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
matiasedgeandnode: done

packages/horizon/contracts/data-service/utilities/ProvisionManager.sol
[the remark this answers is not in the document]
matiasedgeandnode: like the changes, keeping them. Makes the contract smaller, we are slightly above 23kb and would like to keep it there.

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: done

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -670,6 +768,16 @@ contract DisputeManager is
         return subgraphService;
     }
 
+    /**
+     * @notice Get the address of the extended subgraph service
+     * @dev Will revert if the subgraph service is not set
+     * @return The extended subgraph service address
+     */
+    function _getSubgraphServiceExtended() private view returns (ISubgraphServiceExtended) {
```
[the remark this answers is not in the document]
matiasedgeandnode: no longer relevant

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: done

packages/subgraph-service/contracts/libraries/IndexingAgreement.sol
matiasedgeandnode: done

packages/horizon/contracts/payments/collectors/RecurringCollector.sol
matiasedgeandnode: ignore

packages/horizon/contracts/data-service/libraries/DataServiceFeesLib.sol
tmigone: I don't think it's important, but this was not included right?

pcarranzav (commented):
LGTM (with two small comments and pending audit)

packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -588,6 +721,18 @@ contract SubgraphService is
...
+        if (_tokensCollected > 0) {
+            // lock stake as economic security for fees
+            _lockStake(
+                _indexer,
+                _tokensCollected * stakeToFeesRatio,
+                block.timestamp + _disputeManager().getDisputePeriod()
+            );
+        }
+    }
+
```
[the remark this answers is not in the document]
tmigone: I think we should inline this. It's a bit confusing imo, it's a fn only used for IndexingFees but with functionality that's almost the same as QueryFees.

packages/subgraph-service/contracts/libraries/Allocation.sol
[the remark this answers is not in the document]
tmigone: This one still has a mix of fn visibilities (internal/external). IMO it's better to commit to external since it's a linked lib.

packages/subgraph-service/contracts/utilities/Directory.sol
```diff
@@ -67,14 +74,36 @@ abstract contract Directory {
...
+            curation,
+            recurringCollector_
+        );
+    }
+
+    /**
+     * @notice Returns the Recurring Collector contract address
+     * @return The Recurring Collector contract
+     */
+    function recurringCollector() external view returns (IRecurringCollector) {
```
tmigone, line 105: I would change this fn to be consistent with the rest of the directory (i.e using internal fns in the Directory), then in `IndexingAgreemetns.sol` we can do:

```
    function _recurringCollector() private view returns (IRecurringCollector) {
        return Directory(address(this))._recurringCollector();
    }
 ```

instead of

```
    function _directory() private view returns (Directory) {
        return Directory(address(this));
    }
```

File: packages/horizon/contracts/payments/collectors/RecurringCollector.sol [hunk 55]
```diff
@@ -0,0 +1,533 @@
...
+        uint256 collectionStart = _agreementCollectionStartAt(_agreement);
+        require(
+            collectionEnd != collectionStart,
+            RecurringCollectorZeroCollectionSeconds(_agreementId, block.timestamp, uint64(collectionStart))
+        );
+        require(collectionEnd > collectionStart, RecurringCollectorFinalCollectionDone(_agreementId, collectionStart));
+
+        uint256 collectionSeconds = collectionEnd - collectionStart;
+        require(
+            collectionSeconds >= _agreement.minSecondsPerCollection,
```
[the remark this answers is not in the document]
matiasedgeandnode: hmmm.... same issue with payer canceled agreements. I could allow it in either case (elapsed or payer canceled)...

packages/subgraph-service/contracts/utilities/Directory.sol
matiasedgeandnode: I don't think I can call a `Directory` _private/internal_ function from `IndexingAgreements` lib.

matiasedgeandnode:
Final PR [here](https://github.com/graphprotocol/contracts/pull/1217/files)

PR #1140, opened 2025-04-11
Subgraph Service: Integration tests

Maikol:


packages/subgraph-service/test/integration/after-transition-period/subgraph-service/indexer.test.ts
```diff
@@ -0,0 +1,636 @@
...
+        const beforeLockedTokens = await subgraphService.allocationProvisionTracker(indexer.address)
+
+        // Build allocation proof
+        const signature = await generateAllocationProof(allocationPrivateKey, [indexer.address, allocationId])
+
+        // Attempt to create an allocation with the same ID
+        const data = ethers.AbiCoder.defaultAbiCoder().encode(
+          ['bytes32', 'uint256', 'address', 'bytes'],
+          [subgraphDeploymentId, allocationTokens, allocationId, signature],
+        )
```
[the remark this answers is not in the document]
tmigone: then reuse it for all instances in this script

packages/toolshed/src/core/attestations.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  // Encode the receipt using the dispute manager
+  const receiptHash = await disputeManager.encodeReceipt(receipt)
+
+  // Sign the receipt hash with the allocation private key
+  const signature = signer.signingKey.sign(ethers.getBytes(receiptHash))
+  const sig = ethers.Signature.from(signature)
+
+  // Concatenate the bytes directly
+  return ethers.concat([
```
tmigone, line 36: nice, didn't know about ethers.concat utility

packages/toolshed/src/core/attestations.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @param disputeManager The dispute manager contract instance.
+ * @param signer The allocation ID that will be signing the attestation.
+ * @param requestCID The request CID.
+ * @param responseCID The response CID.
+ * @param subgraphDeploymentId The subgraph deployment ID.
+ * @returns The attestation data.
+ */
+export async function createAttestationData(
+  disputeManager: IDisputeManager,
+  signer: Wallet,
```
tmigone, line 16: Let's try to use `HardhatEthersSigner` for any wallet/signer type. It's already what we use in most of the codebase and being consistent will lead to less issue.s

packages/toolshed/src/core/collectors.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param serviceProvider The service provider
+ * @param dataService The data service
+ * @param timestampNs The timestamp in nanoseconds
+ * @param valueAggregate The value aggregate
+ * @param metadata The metadata
+ * @returns The encoded signed RAV calldata
+ */
+export async function getSignedRAVCalldata(
+  graphTallyCollector: IGraphTallyCollector,
+  signer: HDNodeWallet,
```
tmigone, line 20: Same as before, let's use `HardhatEthersSigner`

packages/toolshed/src/core/collectors.ts
```diff
@@ -0,0 +1,84 @@
...
+    collectionId: ethers.zeroPadValue(allocationId, 32),
+    payer: payer,
+    serviceProvider: serviceProvider,
+    dataService: dataService,
+    timestampNs: timestampNs,
+    valueAggregate: valueAggregate,
+    metadata: metadata
+  }
+
+  const encodedRAV = await graphTallyCollector.encodeRAV(ravData)
```
tmigone, line 39: Same as before, I think if we can remove the need for a contract interaction here it's better.

packages/toolshed/src/core/collectors.ts
tmigone, line 1: we should probably rename to `graph-tally.ts` ?

packages/toolshed/src/core/collectors.ts
```diff
@@ -0,0 +1,84 @@
...
+ * @param graphTallyCollector The Graph Tally Collector contract
+ * @param signer The signer
+ * @param chainId The chain ID
+ * @param proofDeadline The deadline for the proof
+ * @param signerPrivateKey The private key of the signer
+ * @returns The encoded signer proof
+ */
+export async function getSignerProof(
+  graphTallyCollector: IGraphTallyCollector,
+  signer: HDNodeWallet,
```
tmigone, line 60: `HardhatEthersSigner`

packages/toolshed/src/core/collectors.ts
```diff
@@ -0,0 +1,84 @@
...
+  proofDeadline: bigint,
+  payer: string
+): Promise<string> {
+  // Create the message hash
+  const messageHash = ethers.keccak256(
+    ethers.solidityPacked(
+      ['uint256', 'address', 'string', 'uint256', 'address'],
+      [
+        chainId,
+        await graphTallyCollector.getAddress(),
```
tmigone, line 71: here we can probably just pass the collector's address as a parameter

packages/toolshed/src/deployments/subgraph-service/actions.ts
```diff
@@ -0,0 +1,62 @@
...
+async function collect(
+  contracts: { SubgraphService: ISubgraphService },
+  signer: HardhatEthersSigner,
+  args: Parameters<ISubgraphService['collect']>,
+): Promise<bigint> {
+  const { SubgraphService } = contracts
+  const [indexer, paymentType, data] = args
+
+  const tx = await SubgraphService.connect(signer).collect(indexer, paymentType, data)
+  const receipt = await tx.wait()
```
tmigone, line 39: we dont need to .wait() here, it's already automatically awaited by the toolshed internal wrappers :)

packages/toolshed/src/deployments/subgraph-service/actions.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const iface = new Interface(['event ServicePaymentCollected(address indexed serviceProvider, uint8 indexed feeType, uint256 tokens)'])
+  const event = receipt.logs.find(log => log.topics[0] === iface.getEvent('ServicePaymentCollected')?.topicHash)
+  if (!event) throw new Error('ServicePaymentCollected event not found')
+
+  return BigInt(event.data)
+}
+
+// Generate allocation proof for the subgraph services
+async function generateAllocationProof(
```
tmigone, line 50: I think we should reserve the "actions" for state modifying interactions with the protocol. We don't need to send a contract call to generate allocation proofs, we could have a helper in `tooolshed/core/allocations`

tmigone: actually we already have `generateAllocationProof()` in `src/core/allocation.ts` 😄

packages/subgraph-service/tasks/test/seed.ts
```diff
@@ -0,0 +1,110 @@
...
+import { indexers } from './fixtures/indexers'
+
+task('test:seed', 'Seed the test environment, must be run after deployment')
+  .setAction(async (_, hre) => {
+    // Get contracts
+    const graph = hre.graph()
+    const { generateAllocationProof } = graph.subgraphService.actions
+    const horizonStaking = graph.horizon.contracts.HorizonStaking as unknown as IHorizonStaking
+    const subgraphService = graph.subgraphService.contracts.SubgraphService as unknown as SubgraphService
+    const disputeManager = graph.subgraphService.contracts.DisputeManager as unknown as DisputeManager
```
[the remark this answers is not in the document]
tmigone: actually you already rebased haha my local branch was not so you should be able to safely remove the type assertions

tmigone (commented):
Looking great, a few suggestions mostly for consistency of the codebase

packages/toolshed/src/deployments/subgraph-service/actions.ts
tmigone: so turns out that the existing `generateAllocationProof()` is for the legacy allocation proof, so we need to add the new variant. Im doing that since i need it for something else.

Maikol: Yes, I was adding `generateSubgraphServiceAllocationProof()` but it was too long so I thought about doing it here. This function I added should be for the subgraph service variant.

tmigone: I added functionality here: https://github.com/graphprotocol/contracts/pull/1116/commits/88b05273505afd85de272f033ea41df565ffa1b5

now we have `encodeAllocationProof()` and `encodeLegacyAllocationProof()` (changed generate to encode to mimic the contracts convention).

tmigone: though now that i think of it, contracts use `encode...` but they dont sign the message, so perhaps `generate...` is more appropriate 🤔

tmigone: yeah lets use `generate...` i'll roll back that change.

tmigone: done in `25624e03`

packages/toolshed/src/core/attestations.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Creates an attestation data for a given request and response CIDs.
+ * @param disputeManager The dispute manager contract instance.
+ * @param signer The allocation ID that will be signing the attestation.
+ * @param requestCID The request CID.
+ * @param responseCID The response CID.
+ * @param subgraphDeploymentId The subgraph deployment ID.
+ * @returns The attestation data.
+ */
+export async function createAttestationData(
```
tmigone, line 14: lets use `generateAttestationData()` for consistency with the other helpers where we return a signed message

packages/toolshed/src/deployments/subgraph-service/actions.ts
Maikol: I'm waiting to get the receipt so I can read the logs for events. Is there another way of getting the receipt?

tmigone: no nvm, that's perfect

packages/subgraph-service/scripts/test/integration
tmigone, line 1: One additional thing, it's not necessary for integration testing but can we still create and run a transfer ownership task that assigns `SubgraphService` and `DisputeManager` ownership?

tmigone (approved):
🎖️

PR #1142, opened 2025-04-23
fix: add extra parameters to contract events (OZ CR-04)

tmigone:


packages/horizon/contracts/interfaces/internal/IHorizonStakingMain.sol
```diff
@@ -189,12 +189,14 @@ interface IHorizonStakingMain {
      * @param verifier The address of the verifier
      * @param delegator The address of the delegator
      * @param tokens The amount of tokens undelegated
+     * @param tokens The amount of shares undelegated
```
Maikol, line 192: should be `@param shares The amount of shares undelegated`

Maikol (approved):

PR #1143, opened 2025-04-23
fix: prevent legacy slash from breaking provision accounting (OZ H-01)

tmigone:

### DIFF
--- a/packages/subgraph-service/contracts/DisputeManager.sol
+++ b/packages/subgraph-service/contracts/DisputeManager.sol
@@ -197,7 +197,7 @@ contract DisputeManager is
     }
 
     /// @inheritdoc IDisputeManager
-    function createLegacyDispute(
+    function createAndAcceptLegacyDispute(
         address allocationId,
         address fisherman,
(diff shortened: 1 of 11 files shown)


Maikol (approved):

packages/subgraph-service/contracts/DisputeManager.sol
```diff
@@ -196,6 +196,45 @@ contract DisputeManager is
         return (dId1, dId2);
     }
 
+    /// @inheritdoc IDisputeManager
+    function createLegacyDispute(
```
pcarranzav, line 200: nit: might be better to call this `slashLegacyDispute` or `createAndAcceptLegacyDispute` to make it clearer that it's not just creating it but also settling it?

pcarranzav (approved):

packages/subgraph-service/contracts/DisputeManager.sol
tmigone: changed to `createAndAcceptLegacyDispute`

PR #1166, opened 2025-05-07
test: add multicall integration tests

tmigone:


packages/horizon/test/integration/after-transition-period/multicall.test.ts
```diff
@@ -0,0 +1,103 @@
+import hre from 'hardhat'
+
+import { ethers } from 'hardhat'
+import { expect } from 'chai'
+import { ONE_MILLION } from '@graphprotocol/toolshed'
+import { setGRTBalance } from '@graphprotocol/toolshed/hardhat'
+
+import type { HardhatEthersSigner } from '@nomicfoundation/hardhat-ethers/signers'
+
+describe('Service Provider', () => {
```
Maikol, line 10: Should we use a specific name here to indicate these tests are for multicall?

packages/horizon/test/integration/during-transition-period/multicall.test.ts
```diff
@@ -0,0 +1,100 @@
+import hre from 'hardhat'
+
+import { ethers } from 'hardhat'
+import { expect } from 'chai'
+import { ONE_MILLION, PaymentTypes } from '@graphprotocol/toolshed'
+import { setGRTBalance } from '@graphprotocol/toolshed/hardhat'
+
+import type { HardhatEthersSigner } from '@nomicfoundation/hardhat-ethers/signers'
+
+describe('Service Provider', () => {
```
Maikol, line 10: Same question as on other test.

Maikol (approved):

PR #1187, opened 2025-06-11
chore: move all interfaces to dedicated package

tmigone:


.gitignore
```diff
@@ -84,3 +84,6 @@ tx-builder-*.json
 **/subgraph-service-localhost/
 **/subgraph-service-hardhat/
 !**/ignition/**/artifacts/
```
RembrandtK, line 86: Can remove `!**/ignition/**/artifacts/`?
Consider `**/*-localhost/` and `**/*-hardhat/` for those that match, should be safe?
I do sometimes wonder if we should add package .gitignores, not try to have all as global patterns. Maybe not so much for this particular case though.

packages/horizon/hardhat.config.ts
```diff
@@ -4,18 +4,19 @@ import '@nomicfoundation/hardhat-toolbox'
 import '@nomicfoundation/hardhat-ignition-ethers'
 import 'hardhat-contract-sizer'
 import 'hardhat-secure-accounts'
+import 'hardhat-graph-protocol'
 
 import { hardhatBaseConfig, isProjectBuilt, loadTasks } from '@graphprotocol/toolshed/hardhat'
 import type { HardhatUserConfig } from 'hardhat/types'
 
-// Skip importing hardhat-graph-protocol when building the project, it has circular dependency
+// Some tasks need compiled artifacts to run so we avoid loading them when building the project
```
RembrandtK, line 12: I would try to change package structure to remove the dependency. I suspect runtime loading can be eliminated, and it is better to eliminate.

packages/interfaces/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+import { IGraphPayments } from "@graphprotocol/interfaces/contracts/horizon/IGraphPayments.sol";
+
+contract GraphPayments is IGraphPayments {
+    ...
+}
+```
+
+Note that contracts in the `toolshed/` directory are not meant to be imported by Solidity code, they only exist to
+generate complete TypeScript types.
```
RembrandtK, line 32: I noticed them. Not looked at why, I presume to resolve something it was hard to otherwise do?

packages/interfaces/contracts/toolshed/IRewardsManagerToolshed.sol
RembrandtK, line 1: General observation, sorry if painful to implement. :) Would a suffix like `Type` be better than `Toolshed`? (Make it about the purpose rather than where introduced?)

packages/interfaces/hardhat.config.ts
RembrandtK, line 1: This might be a problem you inherited from the way I set things up initially, if you adopted some of my config: locally my common does not, although it did at one stage, compile locally. The local compilation causes knock-on issues. Does it also contribute to needing for Toolshed interfaces?

.gitignore
tmigone: - Removed the artifacts rule
- Switched to
	```
	**/ignition/deployments/*-localhost/
	**/ignition/deployments/*-hardhat/
	**/ignition/deployments/*-31337/
	**/ignition/deployments/*-1337/
	```

packages/horizon/contracts/libraries/LinkedList.sol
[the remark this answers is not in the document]
tmigone: Sorry not sure I follow here. Avoid creating what?

packages/interfaces/README.md
tmigone: The reason is interfaces are not 1:1 with contract implementations. Storage variables, dependencies without interface files are some examples of things that add external functions to an implementation but are not usually in the interface files. The toolshed stuff (very bad name btw) is supposed to bridge that gap by adding back anything that's missing. 
You should be able to include everything into the actual contract interface that gets imported and usually extended by the implementation, but I didn't want to mess with audited code at this point.

packages/interfaces/contracts/toolshed/IRewardsManagerToolshed.sol
tmigone: yes to all. It would be better and slightly painful, but worth it I believe. I'll add it to my to-do list, but need to knock off some higher prio items first.

package.json
```diff
@@ -10,7 +10,7 @@
     "postinstall": "husky",
     "clean": "pnpm -r run clean",
     "clean:all": "pnpm clean && rm -rf node_modules packages/*/node_modules",
-    "build": "chmod +x ./scripts/build && ./scripts/build",
+    "build": "pnpm -r run build",
```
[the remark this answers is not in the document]
RembrandtK: I think we need child packages, and/or siblings that combine dependencies across contract packages.

They are to break cyclic dependencies, I think interfaces does not by itself fully solve that. It might mean that contract packages no longer rely on toolshed? Or is that still via dynamic loading? However, deployment has cross package needs for compiled contracts. (I think there is an argument for having a single deploy sibling, that spans contract packages.)

The same scenario for tests, similar nuances to deploy, but I think not as strong a case for (just) a single test package spanning contract packages.

If we do not separate them, I think we can appear to get away with it by recompiling dependencies, which the hardhat-dependency-compiler can do, but then you end up with other issues due to that duplication and inconsistency.

I suspect contract compilation should be separate reusable step from what we do with the compilied artifacts (deployment, testing, etc). But I might be missing something; I am new to this package management system.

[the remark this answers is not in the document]
tmigone: Got it. I'm not too sold on any alternative to be honest, there are things i dislike in both :p. But happy to keep iterating.

packages/horizon/contracts/libraries/LinkedList.sol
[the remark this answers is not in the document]
tmigone: Ah got it. So back then we looked at the available options and there was nothing that was exactly what we needed. I don't remember top of my head what exactly was missing from OZ's EnumerableSet implementation but it didn't quite fit the bill.

packages/interfaces/hardhat.config.ts
tmigone: I think we can consider types to be compiler agnostic. They only use the ABIs from the compiled artifacts, which shouldn't change much between different compiler versions. Even if a new compiler version added a new feature that would change the ABI significantly, the previous ABI would still be compatible and useful for typings. I think it's unlikely we hit any problems related to this so I wouldn't add extra complexity for now. Also I suspect always compiling with the latest possible compiler would produce ABIs that are universal.

RembrandtK: Sounds reasonable. We could generate ABIs without compilation?

PR #1206, opened 2025-07-24
[audit-09] fix: [TRST-L-6] Proper agreement version check

matiasedgeandnode:
Screenshot 2025-07-24 at 10 47 35


packages/subgraph-service/test/unit/subgraphService/indexing-agreement/accept.t.sol
```diff
@@ -226,7 +226,38 @@ contract SubgraphServiceIndexingAgreementAcceptTest is SubgraphServiceIndexingAg
         subgraphService.acceptIndexingAgreement(ctx.indexers[0].allocationId, accepted);
     }
 
-    function test_SubgraphService_AcceptIndexingAgreement_Revert_WhenAgreementAlreadyAllocated() public {}
+    function test_SubgraphService_AcceptIndexingAgreement_Revert_WhenAgreementAlreadyAllocated(
```
pcarranzav, line 229: This test seems unrelated to the fix, right? In which case it would be good to include a test for when an invalid version is supplied

pcarranzav (approved):

PR #1208, opened 2025-07-24
fix: [TRST-L-9] Cancel agreement if over-allocated

matiasedgeandnode:
Screenshot 2025-07-24 at 14 15 27


packages/subgraph-service/contracts/SubgraphService.sol
```diff
@@ -737,7 +737,20 @@ contract SubgraphService is
...
+
+        (uint256 paymentCollected, bool allocationForceClosed) = _presentPOI(
+            allocationId,
+            poi_,
+            poiMetadata_,
+            _delegationRatio,
+            paymentsDestination[_indexer]
+        );
+
+        if (allocationForceClosed) {
```
Maikol, line 749: Another option would be checking the allocation storage for the state after calling `presentPOI` it's slightly more expensive but we save on contract size, though I'm not sure where we're at with the new libraries.

Maikol (approved):
Looks good to me.

pcarranzav (approved):

packages/subgraph-service/contracts/SubgraphService.sol
matiasedgeandnode: SubgraphService is ~23.4kb - So rather close.

matiasedgeandnode:
Final PR [here](https://github.com/graphprotocol/contracts/pull/1217/files)

PR #1221, opened 2025-09-17
Rewards Eligibility Oracle (ERO)

RembrandtK:
WIP and expected to be re-based onto main after Horizon has been merged.


pcarranzav:
nit: isn't it REO rather than ERO?

PR #1261, opened 2025-11-28
feat: post horizon transition period clean up

tmigone:
This PR removes code that's no longer relevant after the transition period. 

In order for these changes to be valid there are some assumptions about the network state that must hold:
1) **There are no more legacy allocations open.** This can be guaranteed by force closing allocations, which can be done permisionlessly 28 days after Horizon go live.
2) **Legacy allocation ids have been fully migrated to the `SubgraphService`.** This is a privileged account operation that will be executed once Horizon is live and legacy allocations can no longer be created.
3) **All service provider legacy unstake operations must be fully thawed.** This needs to be monitored but it's also guaranteed as long as the transition period is longer than 28 days.

---

Note that `contracts` package tests are failing due to removing the staking contract as a valid rewards issuer from the `RewardsManager`. A deeper refactor of the package is required to fix this, most contracts in the `contracts` package are now deprecated so it's likely we need to extract those that are not into `horizon` package or a brand new one.


packages/horizon/contracts/staking/HorizonStaking.sol
```diff
@@ -616,54 +547,25 @@ contract HorizonStaking is HorizonStakingBase, IHorizonStakingMain {
...
     }
 
     /**
      * @notice Withdraw service provider tokens once the thawing period (initiated by {unstake}) has passed.
      * All thawed tokens are withdrawn.
-     * @dev TRANSITION PERIOD: This is only needed during the transition period while we still have
-     * a global lock. After that, unstake() will automatically withdraw.
+     * This function is for backwards compatibility with the legacy staking contract.
+     * It only allows withdrawing tokens unstaked before horizon upgrade.
+     * @dev This function can't be removed in case there are still pre-horizon unstakes.
```
RembrandtK: So presumed in this scenario they are no longer locked although not yet withdrawn?

RembrandtK:
The use of __DEPRECATED prefix might be misleading, because although no longer intended to be the main flow, in many cases the variables are still being used and have valid values that can still be updated. Not sure if or how to change, it might be another prefix (__LEGACY) could be used for this scenario? You could argue 'deprecated' does mean this, but I think for the rest of the code it generally implies 'no longer used', representing a field that could (for example) be repurposed?

Maikol (commented):
We should also delete the files in `subgraphService/test/integration/during-transition-period/`

Maikol (approved):
Looks good to me!

RembrandtK:
Incorporated into: https://github.com/graphprotocol/contracts/pull/1301

Not expected to be directly merged. Can be closed once audit https://github.com/graphprotocol/contracts/pull/1301 is merged after audit completes.

PR #1301, opened 2026-03-01
Indexing payments management audit

RembrandtK:
Filter for files in audit scope:

```
git diff main --name-status --diff-filter=d -- \
    '*.sol' ':!*/test*' ':!*.t.sol' ':!*Helper.sol' ':!*/mock*' ':!*/toolshed/*'`
```

bf6d4cb70 Indexing payments (merge)
Previously audited (2025-06-Indexing-Payments.pdf) implementation of indexing agreements for the Graph Protocol's recurring payment system. Adds RecurringCollector, IndexingAgreement library, and integrates indexing fee collection into SubgraphService and DisputeManager.

28edcd7a3 Post-Horizon migration removal of transition code (merge)
Remove HorizonStakingExtension, ExponentialRebates, LibFixedMath, MathUtils, and all transition-period guard logic. Simplify HorizonStaking to direct implementation without delegatecall split. Clean up RewardsManager multi-issuer model to single subgraphService. Remove legacy allocation/dispute code from SubgraphService.

f4451f189 feat: add back legacy allocation id collision check
fa9951427 fix: cap maxSecondsPerCollection instead of reverting
c6836a716 fix: enforce temporal validation on zero-token collections and remove zero-POI special case
8efaec97d feat: add adjustThaw to PaymentsEscrow
3f1578cdc refactor: rename IRewardsEligibility to IProviderEligibility
(For the purpose of this audit note that the interfaces being changed are not in production.)

d20bc844d feat: contract approver model for RecurringCollector accept/update
ec7236086 feat: IDataServiceAgreements interface and SubgraphService integration
89def3d34 feat: enumerable indexer tracking for REO and issuance constructor cleanup
(Do not need storage compatibility for delpoying an upgrade.)

843fdd2d0 feat: RecurringAgreementManager with lifecycle, escrow funding, and agreement updates

PR #1302, opened 2026-03-02
Fix underflow in _updateSubgraphRewards

RembrandtK:
fix(rewards): reorder subtraction in _updateSubgraphRewards to avoid underflow
The old code's two hooks wrote to different fields with inverted semantics:
- onSubgraphSignalUpdate set accRewardsForSubgraph (A) from storage
- onSubgraphAllocationUpdate set accRewardsForSubgraphSnapshot (S) from
  a view (storage + pending), so S leads and A lags after allocation updates

After the proxy upgrade, _updateSubgraphRewards computed A.sub(S).add(P)
which underflows on the intermediate A - S when A < S. Rearranging to
A.add(P).sub(S) adds pending rewards first, avoiding the intermediate
underflow. S <= A + P always holds because P covers T1→now (a superset
of the T1→T2 gap S - A).

Observed on Arbitrum Sepolia: A < S by ~7,235 GRT for subgraphs whose
last pre-upgrade interaction was onSubgraphAllocationUpdate. All reward
operations (signal, allocation, claim) reverted permanently.

### DIFF
--- a/packages/contracts-test/tests/unit/rewards/rewards-snapshot-inversion.test.ts
+++ b/packages/contracts-test/tests/unit/rewards/rewards-snapshot-inversion.test.ts
@@ -1,0 +1,436 @@
+import { Curation } from '@graphprotocol/contracts'
+import { GraphToken } from '@graphprotocol/contracts'
+import { IStaking } from '@graphprotocol/contracts'
+import { RewardsManager } from '@graphprotocol/contracts'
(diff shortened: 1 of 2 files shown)


tmigone (approved):

RembrandtK:
To be merged with (when audit complet): https://github.com/graphprotocol/contracts/pull/1301

PR #1321, opened 2026-04-17
chore: deduplicate zero-collection test

Olexandr88:
Removed test case for collecting zero tokens.


RembrandtK:
Thanks for the cleanup effort! However, these two tests aren't actually duplicates — they cover different scenarios:

- The one at **line 636** runs inside a parameterized loop with delegation and varied fee/tax configurations (4 combinations).
- The one at **line 772** (removed here) runs with default parameters and no delegation.

Since they exercise different code paths through the rebate and fee distribution logic, I think we should keep both. Closing in favour of keeping the existing coverage.

PR #1327, opened 2026-04-28
chore: set minimum release age on pnpm to 5 days

tmigone:


RembrandtK (dismissed):
Might want to consider adding something like following so they do not block internal propagation:

```
minimumReleaseAgeExclude:
  - '@graphprotocol/*'
  - hardhat-graph-protocol
```

RembrandtK (dismissed):

RembrandtK (approved):

